PR #30, opened 2014-12-26
Add OSX/Win support for Piranha xtreme USB gamepad.

stolk:


AliceLR:
Just came across a Snakebyte controller that has the same GUID in Windows as this controller does in the latest SDL (`030000008f0e00000300000000000000`). However, the trigger and shoulder buttons are swapped on both sides (i.e. each shoulder button presses the trigger and each trigger presses the shoulder button).  In Linux, this controller is detected but without this issue, and in Mac OS X it is detected with a different name/mapping from the Mac mapping in this PR (and it doesn't have the issue).

Can you confirm that the GUID pasted above is correct for the Piranha controller in newer SDL versions or give more info about the controller?

stolk:
SDL2 Gamepad tool says:

_SDL2 Gamepad Tool v1.2 by General Arcade (compiled with SDL version 2.0.7, DLL version 2.0.7)
Website: http://generealarcade.com/gamepadtool/

Searching gamepads...
Found 1 gamepad(s):
"Piranha xtreme", 030000008f0e00000300000000000000 (mapping available)
Environment variable "SDL_GAMECONTROLLERCONFIG" is not defined
Checking if new mappings available from github: https://github.com/gabomdq/SDL_GameControllerDB_

And the triggers/bumpers are correct.

AliceLR:
Thanks for checking! I'm not surprised the problem is with this Snakebyte controller (it's not the only thing wrong with it...)

PR #32, opened 2015-01-31
Added HORI Gem Pad 3 to OSX mappings.

boondaburrah:
Mapped using big picture mode on steam. Controller is here: http://stores.horiusa.com/playstation-3-gem-pad-ruby/ (standard USB HID joystick with Playstation labels. No rumble)


gabomdq:
Merged manually due to lack of "platform" field. Thanks!

PR #34, opened 2015-02-22
Added "Microsoft X-Box 360 pad" on Linux

tomsik68:
Mappings for the gamepad were created using antimicro.


Nixola:
Weird. The gamepad worked for me out of the box, as soon as I plugged it in.

PR #52, opened 2015-10-01
Improved Wii( Re)mode mapping for Mac OS X

LRFLEW:
To clarify the changes I've made:

I've renamed it from "Wiimote" to "Wii Remote". This is what Nintendo officially calls it; "Wiimote" is just a nickname. Since the other controllers use their official names, it makes sense to change this one to it as well.

The other change I made was in the mapping. `b11` is the C button and `b12` is the Z button. Since the C button is above the Z button, it would make sense to have it mapped to RB. However, the mapping currently has the C button mapped to RT, which doesn't make a lot of sense. I switch the two around so that it's in the more natural order.


gabomdq:
@corsix You contributed the current mapping in the database, what's your take on these proposed changes?

corsix:
Seems reasonable.

PR #55, opened 2015-11-01
Update gamecontrollerdb.txt

Somnium42:
Add Support for the Xeox Pro gamepad.
Wii u Pro controller A-B-X-Y button order fixed.


gabomdq:
@joaorb64 what's your take on the proposed Wii U Pro changes?

joaorb64:
Game controllers have all kinds of patterns, numbers or letters on their buttons. But I personally think a game developer designs a game thinking of their actual layout, their position.
The Xbox controller buttons have the same letters as Nintendo's, but inverted. So it's a choice of whether you want the right layout or the right on-screen button letter.

gabomdq:
Added the XEOX Pro gamepad.

PR #62, opened 2016-02-13
add linux mapping for "nacon" x360 clone

Hinsbart:


gabomdq:
Added manually since this PR didn't include the platform name.

PR #63, opened 2016-03-02
GameCube {HuiJia USB box}

mathewv:
Just a config for the USB GameCube controller adapter I own. I tried to make the name/layout match the existing GameCube controller config.


gabomdq:
Merged manually due to lack of platform specification in PR.

PR #80, opened 2016-06-12
Added SFC30 Joystick

gingerbeardman:
Mac OS X, USB wired mode.

When in this mode it is reported to the system as: SFC Joystick

http://www.8bitdo.com/sfc30/


jmtd:
Just to confirm this, I get the same for this controller in wired mode. I also have a string for Bluetooth controller mode which I'll put in another PR

PR #84, opened 2016-06-20
Update gamecontrollerdb.txt

mariuszk911:
8Bitdo SFC30 GamePad Bluetooth, fw 2.68

PR #87, opened 2016-06-27
Add controller strings for SFC30 (three platforms)

jmtd:
SFC30 controller: http://www.8bitdo.com/sfc30/

The SFC30 controller can present itself in a variety of modes and it offers up
different names in each. This patch captures data for one mode on three
platforms (Mac OS X, Windows, Linux).

(USB mode on Linux and Windows is missing as the button events did not make it
through to SDL's controllermap tool on Fedora 24/Linux 4.5.5 nor Steam Big
Picture mode on Windows. The two Bluetooth modes were indistinguishable on
Windows.)

There exists a similar controller called the SNES30 (And some others) that are
very likely identical except for the name, but I have not verified this yet so
haven't synthesized lines for those controllers until I can.

This patch incorporates the changes made in #84 (thanks @mariuszk911 which are
correct but not comprehensive.  (I could've just supplied the missing ones but
each of these PRs are going to conflict anyway)


gingerbeardman:
Two of the Mac OS X settings are duplicates

There are only two modes where this works with SDL: USB and Bluetooth. 

Do we want to use official default names or not? I can get those too, if that would help. 

I'm happy to do all platforms if that would also help?

gingerbeardman:
USB is wired mode, automatically switches into that when a cable is attached. 

Bluetooth is default, mode 1, switching on without any buttons held. 

Now I think about it, perhaps there's also a "player 2" mode for at least Bluetooth.

jmtd:
If the name field is advisory and doesn't matter then we can dedupe one entry (for another project the name is matched exactly and so I've had to code in the two variants)

Switching on with "R" held is an undocumented bluetooth mode where it presents as a straight controller-  the original firmwares at least it pretended to be a keyboard to work better on Android and/or iOS tablets. For some firmwares the "R" mode is the only BT one that worked for the SDL2 apps I was looking at on OS X, but it's a bit of a moving target with the firmwares.

jmtd:
I've rebased on top of the merged #80 and removed the duplicate lines. I'd still like to double check that the names are advisory only, otherwise a controller could fail if a user had it in a mode which presented a different name to what is recorded.

gabomdq:
Added the new mappings manually, the name of the controller is illustrative only, apps are free to use it for displaying to the user or just ignore them.

jmtd:
Well uh, thanks, I guess.

PR #96, opened 2016-07-06
new 4 game controllers

D0NM:
+ 8Bitdo NES30 PRO Wireless
+ 8Bitdo NES30 PRO USB
+ EXEQ EQ-PCU-02100
+ Gembird JPD-DualForce 
 note: its GUID equals to ff113133000000000000504944564944,SVEN X-PAD 
but SVEN X-PAD's A,B,X,Y buttons mapping is wrong. 
Also it is missing 2 stick buttons b10 and b11. I've added them into Gembird JPD-DualForce mapping.


gabomdq:
Added manually to fix errors in mappings (lines must end with a comma)

PR #97, opened 2016-07-21
added Linux controllers

cprn:
Done with `controllermap` from `SDL2-2.0.4` and tested, works fine.


cprn:
Thanks for pulling. Cheers!

PR #99, opened 2016-08-14
Use axes for triggers (PS3 Controller)

Arvamer:
PS3 controller for most buttons generate two events - button pressed and axis changed. For triggers there is more sense to use axis event.


p-groarke:
All the playstation mappings have been updated. Could you rebase and reopen with updated changes, including the other Playstation mappings? Thx!

Arvamer:
I no longer have PS3 controller and I don't want to do this without testing.

PR #107, opened 2016-10-13
Correct and new DS4 mappings

soeb:
This PR includes new mappings (DS4 v2 on Linux via usb/bt, DS4 Adaptor on Windows and on Linux) and corrections for Windows and Linux DS4 mappings (axes instead of buttons for triggers, not using share button for back).


johndrinkwater:
👍

KainXVIII:
Can you add DS4 v2  (cuh-zct2e) mapping for Windows?

PR #111, opened 2016-11-13
Add mappings for F310 (DInput) on Linux

jgkamat:
Adding mappings for the Logitech F310 Controller on Linux when in DInput Mode.


jgkamat:
Thanks! :smile:

PR #117, opened 2017-01-14
Add Raphnet USB Sega Saturn adaptor

mistydemeo:
This adds a mapping for the [Raphnet USB Sega Saturn adaptor](http://www.raphnet.net/electronique/saturn_usb/index.php).

Is it valid to specify both axis and button mappings for something? This adaptor supports both the digital gamepad and the 3D gamepad; the latter has analogue triggers, while the former has digital buttons. I've tried mapping it twice here, but I don't know if that's valid.


gabomdq:
I think this might be a valid use case but I haven't verified it.

mistydemeo:
Had a chance to test this - seems to be working fine. Thanks for merging!

PR #119, opened 2017-02-07
update gamecontrollerdb.txt - linux psx usb

kerpal-j:
Bus 001 Device 049: ID 6666:0667 Prototype product Vendor ID WiseGroup Smart Joy PSX, PS-PC Smart JoyPad

it's an ancient ps1 -> usb adapter


p-groarke:
Added, thanks for the PR!

PR #124, opened 2017-03-12
Add Microsoft Xbox 360 Wireless controller

adler187:
Official Microsoft Xbox 360 Wireless controller dongle on Linux.


cgutman:
This mapping is required after the Linux xpad driver was updated in Linux v4.10 to report the same product ID for wireless Xbox 360 gamepads that Windows does. The commit went to the stable queue as well, so it is being backported to LTS and stable kernels.

p-groarke:
Added, thanks for the PR!

PR #131, opened 2017-05-05
Add support for the NGS Phantom USB gamepad in Linux

Swyter:
@gabomdq It's the same gamepad from  #116 but with the Linux-specific mapping. Had to change the analog axes, but the digital button mappings are still the same.

imagen


p-groarke:
Added as DragonRise. Please re-open if the mappings aren't working.

PR #132, opened 2017-05-09
added different version of Sony DualShock 4 (Linux)

oncer:
For some reason, my DS4 was not being recognized, it has a slightly
different GUID.
I copied the mapping from Steam big picture mode.


CookiePLMonster:
It's most likely DS4 v2 from PS4 Slim?

oncer:
I bought the controller in late 2014, so I don't think it belongs to the PS4 slim. I have 4 of those, and I just checked, they all have the same GUID (030000004c050000c405000011810000).

mathieu-lemay:
I also have two of these controllers. Got them both in early 2014, so definitely not PS4 Slim.

p-groarke:
Added, thanks for the PR!

Issue #110, opened 2016-11-11
New GUID format in SDL 2.0.5


slouken:
In SDL 2.0.5 the USB device GUIDs are standardized across platforms to the Linux format.

For example, the Windows GUID gets a 03 prefix (USB) and the VID and PID are split and the final string is removed:
341a3608000000000000504944564944
becomes
03000000341a00003608000000000000

On Mac OS X the device version has been added to the GUID and it has been split similarly:
4c05000000000000c405000000000000
becomes
030000004c050000c405000000010000

slouken:
I added code to support older mapping GUIDs, so this isn't urgent, but FYI.
https://hg.libsdl.org/SDL/rev/20855a38e048

gabomdq:
Hey Sam, how do you think we should handle this change here? It's possible now that if people generate mappings with SDL 2.0.5, they won't work on previous versions...but they will work the other way around...should I only accept old style GUIDs? Split the db in two separate files for pre and post 2.0.5?

slouken:
Maybe split the file? Or remap the guids to the older format before you accept them?

ticky:
Oh, is _this_ why Steam Big Picture’s mappings ended up with different UUIDs to when I used controllermap? Good to know!

Is there a way to convert mappings? I presume the GUID discards enough information that it’s impossible…

ticky:
Oh huh never mind, just read your patch, haha!

Maybe SDL_GameControllerDB should normalise to one or both, yeah!

eddieray:
So evidently, v2.0.5 doesn't actually support the older mappings?

ticky:
SDL v2.0.5 _is_ backwards-compatible with the old mapping types, as mentioned in [this comment](https://github.com/gabomdq/SDL_GameControllerDB/issues/110#issuecomment-259952120).

eddieray:
v2.0.5 was released 6 months ago... that "fix" was made 4 months ago. :-(

ticky:
Ah, my mistake! Yeah, that’s awkward.

eddieray:
The fix seems doable right in my game code on Windows if I read the file and convert the GUIDs before adding them to SDL... then I can use them with the v2.0.5 released version of SDL.  Or maybe it'll be easier to just convert the file in advance.

[...]

PR #133, opened 2017-05-12
Add Twin USB Joystick mapping for Android

NY00123:
Note that this is the same adapter currently identified in the database as "Twin USB PS2 Adapter" on Linux, although the button numbers differ (for technical reasons specific to Android).

Further note that this mapping's GUID is given in the older format (compatible with SDL v2.0.0-2.0.4).

BTW, it looks like one of Travis' checks fails, since "Android" isn't currently recognized as a proper platform value.


p-groarke:
Android support has been added with a few initial controllers. Could you update your PR with the new changes so I can pull?

p-groarke:
Your mapping has been merged. Thank you for the pr.

NY00123:
Hey there,

Thanks for merging the mapping (and further for your other commits and mergers)!

I shall probably say the mapping is still using the v2.0.4-compatible GUID format, although I'm not really sure there was any change to the format on Android.

Also, as mentioned in #110, code was added in order to support the older formats in v2.0.5+, although it's specific to WIN32 and macOS (so again probably irrelevant to Android): https://hg.libsdl.org/SDL/rev/20855a38e048

Thus, I dunno if it the above mapping shall rather reside in older_releases/gamecontrollerdb_204.txt, assuming this file is ever updated at all.

Note that I haven't really checked the controller on Android for a while, but hopefully it's not necessary to re-generate any GUID.

p-groarke:
Hey! No problem, I'm trying to revive this thing ;)

It is true the GUID isn't in the 2.0.5 format, but as much can be said for the SDL header mappings, which are the "golden standard" and official. Since SDL hasn't updated their GUIDs, I'll assume it is ok? If you want to test on the latest SDL version and send the mapping with a different GUID, I'll happily accept it.

The 204 file is in maintenance mode, I know of at least 1 engine still using that format (Love2D), and I'm guessing others are as well. It's not my focus, but there is no reason not to provide fixes when appropriate :) Adding a new platform might be crossing the line though. You are free to convince me, I don't have strong feelings about that :)

NY00123:
I've just checked, and looks like there was no change to the GUID on platforms other than Windows and macOS (with the possible exception of something to do with inpid.version on Linux, and even then, it does not apply to Android): https://hg.libsdl.org/SDL/rev/9da2701eeb4a

Also, on the contrary to what you've just said, I think SDL did update their GUIDs (as shown in the above commit)? As already known, though, the older GUIDs should still work in v2.0.5+ (or at least in 2.0.6+, based on https://github.com/gabomdq/SDL_GameControllerDB/issues/110#issuecomment-298913634).

I suppose the 2.0.4 db can still be updated, *if* there's any engine depending on the older format while also supporting Android (apparently Love2D is an example); Although, it's true that game controllers are probably less commonly used on Android, and I'm not even sure if the Android-specific mappings are supported by Love2D at all.

p-groarke:
Awesome! Then that's that. I meant SDL updating the Android mappings (not windows and osx). Good day, keep the prs coming ;)

PR #136, opened 2017-06-12
Add NVIDIA Shield Controller support - firmware version 1.04

VenKamikaze:


p-groarke:
Added, thanks for the PR!

PR #137, opened 2017-06-15
add Twin USB for Mac

firodj:


p-groarke:
Added, thanks for the PR!

PR #139, opened 2017-06-21
Add (another) PS4 controller

svanheulen:
I just got a new PS4 controller and it has a different ID then any of the PS4 controllers in the configuration. I also included the model number.


p-groarke:
Added (and many more), thanks for the PR!

PR #141, opened 2017-07-03
Add Xbox One S Controller over Bluetooth, in Linux

duganchen:


p-groarke:
Added, thanks for the PR.

PR #143, opened 2017-07-09
Add: Hori Real Arcade Pro EX mapping, Linux platform

crisisinaptica:


p-groarke:
Mappings have been added. Thanks for the PR!

PR #144, opened 2017-07-13
Added Nintendo Retrolink USB Super SNES Classic Controller

hsjunnesson:


p-groarke:
Added, thanks for the PR!

PR #145, opened 2017-07-23
Add DualShock 4 Wireless Adaptor mapping for Mac OS X.

logicallysound:
Like in the Windows mappings, there's a different GUID for using the wireless adaptor versus connecting the controller directly by Bluetooth.


p-groarke:
Mapping has been added. Thanks for the PR!

PR #149, opened 2017-08-16
Add additional entries from hrydgard/ppsspp

VVVFO:
PPSSPP is trying to use this repository as a submodule for this file, but a few entries in the old file is not present in this repo


p-groarke:
Mapping have been added. Thanks for the PR!

PR #150, opened 2017-08-22
Added Saitek P-Something gamepad.

Lyberta:
Pretty old one but very usable.


p-groarke:
Mapping has been added. Thanks for the PR!

PR #153, opened 2017-09-03
Adding message when the script runs successfully

Thunderforge:
For a while, I thought the script wasn't working correctly because there was no output. This should eliminate that confusion for future users by making it clear that it did run correctly, but found nothing.


p-groarke:

PR #155, opened 2017-10-07
Add Raphnet.net's 4NES4SNES

hackerb9:
Plug in four original Nintendo Famicom or Super Nintendo controllers and
convert them to USB.  (Arduino source code is available at raphnet.net.)


p-groarke:

PR #159, opened 2017-10-14
parse.py improvements

Akaricchi:
* Made it compatible with Python 3, and explicitly specified the Python 3 interpreter. The script could still work with Python 2, but it uses the Python 3 print semantics, and thus should at least use `from __future__ import print_function`  
However, I don't think Python 2 needs to be supported. It's a legacy language, and Python 3 is easily available on any system these days.

* Added a new check: consequent spaces in the name field. A few devices have names with unnecessary spaces with them. It's inconsistent and sometimes gets as bad as 5+ spaces in a row. parse.py will now warn about such names, and automatically fix them when -o is used.

* parse.py will now count errors, and exit with error code 1 if any are detected.


Akaricchi:
Oops, looks like I submitted this to the wrong repo.

PR #164, opened 2017-12-17
Add Mac support for another Tomee SNES USB Controller

andrewhavens:
Generated mapping string using the official SDL 2.0.7 utility.


p-groarke:
Thank you.

PR #170, opened 2018-01-07
Correct hats for X-Box One pads on Linux

### LINKED ISSUE #126: Issue with Microsoft X-Box One padv2
The configuration for the Xbox one pad is wrong there are unnecessary assignation that create problems, here is a corrected version.
030000005e040000dd02000003020000,Microsoft X-Box One padv2,x:b2,a:b0,b:b1,y:b3,back:b6,guide:b8,start:b7,dpleft:h0.8,dpdown:h0.4,dpright:h0.2,dpup:h0.1,leftshoulder:b4,lefttrigger:a2,rightshoulder:b5,righttrigger:a5,leftstick:b9,rightstick:b10,leftx:a0,lefty:a1,rightx:a3,righty:a4,platform:Linux,

Celti:
The current mappings for `X-Box One pad` and `... v2` on Linux have duplicate/spurious `h0.0` hat mappings for *all* d-pad directions as well as the left shoulder button. This causes issues on most games using these. The corrected entries have been tested in gameplay on the Linux (eON) ports of Saints Row 2, Saints Row the Third, and Bioshock Infinite, all seeming to work flawlessly.

This fixes #126, which was seemingly closed due to lack of response.


p-groarke:
I've added tests to check for duplicate keys, so this won't happen again. It required some refactoring, so now keys are alphabetically sorted as well. Should make it easier to spot such mistakes.

Many mappings were affected with the same issue (dpdown, dpright, dpup and leftshoulder set to h0.0). I'm guessing there was a bug in an old tool used to generate mappings. Thanks for reporting, debugging and fixing this issue!

PR #177, opened 2018-01-11
Add new mapping for my old controller

porky11:
Not sure if they are sorted, I just created a keymapping and added it to Linux. It works correctly for my controller, I already tested it.


p-groarke:
I sorted the db. Thanks for the PR.

PR #181, opened 2018-01-16
check.py: fix initialization of Mapping.__keys

Akaricchi:
The code was being interpreted as `"+righty-leftx": ""`


p-groarke:
Oops. Thx for the fix!

PR #183, opened 2018-01-19
Add Art-Tech 4ch R/C Flight Sim controller

zagor:
This is a usb-connected R/C trainer controller made by the chinese company art-tech.com.

Image of Art-Tech 4CH R/C Flight Simulator


p-groarke:
Hmmm, I'm not convinced the mapping is complete enough to include it in the db. Joysticks and specialized controllers use the SDL joystick api, not gamepad. It is more appropriate for this type of hardware, as cool as it is :)

PR #184, opened 2018-01-19
Added mappings for "Dual Box WII" controller

gamezer0:
This adds support for Wii controllers connected through Mayflash' Wii Classic Controller Adapter (http://www.mayflash.com/products/pcusb/pc052.html) like the Classic Controller itself or the NES Mini's controller. I created the mappings using the Wii Classic Controller Goldeneye edition. The mappings were done following the Xbox 360 layout, instead of mapping the name of the buttons (i.e. Xbox 360's A corresponds to Wii Classic Controller's B). Please let me know if it should be done the other way round.

image


p-groarke:
Hello, the GUID is in SDL 2.0.4 format. Could you use either the SDL 2.0.7 controllermap or Steam to update the GUID (no need to redo the full mapping). Thank you!

gamezer0:
Fixed the GUID format and added the mapping to the 2.0.4 DB (I don't need it, but in case it's useful to someone else)

Thanks!

PS: I've seen you've added some mapping examples, that's super helpful! But for a moment I thought I skipped that section the other day and was facepalming myself :P

p-groarke:
Haha it's new indeed. I've been postponing making the guide, but your comment gave me the extra motivation to make the graphics. PR looking good :)

PR #185, opened 2018-01-20
Added entries for Logitech Dual Action.

r-lind:


p-groarke:
Hi, this is a duplicate.

If there are problems with the original mapping, please open an issue detailing the problem. Thx for the pr.

```
Duplicate detected at line #179
Previous mapping at line #178
In mapping
030000006d04000016c2000000000000,Logitech F310 Gamepad (DInput),a:b1,b:b2,back:b8,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,leftshoulder:b4,leftstick:b10,lefttrigger:b6,leftx:a0,lefty:a1,rightshoulder:b5,rightstick:b11,righttrigger:b7,rightx:a2,righty:a3,start:b9,x:b0,y:b3,platform:Mac OS X,
```

p-groarke:
Can you remove the conflicting mapping (030000006d04000016c2000000000000).

p-groarke:
And the edited 030000006d04000016c2000000030000 mapping. This mapping comes from the SDL header, which is the be all end all of mappings. If there are issues with this mapping (030000006d04000016c2000000030000, Mac OS X), you would have to send a patch to the SDL project itself. Thank you for this awesome PR btw!

r-lind:
Sorry if I have gotten this wrong, but I just cut and pasted all version of the Logitech Dual Action to Windows, macOS and Linux, and added the missing v2.0. I fail to see why Travis is complaining.

p-groarke:
No problem. So you cannot assume Windows, Linux and Mac OS X have the same GUID. So just add mappings for the platform you tested (macOS in this case). You can create a new PR if it is easier (dealing with git).
 If you only have 1 new mapping, just add that one (without changing the others). Lets take it from there, I'll point you in the right direction if there are other problems :)

p-groarke:
The GUID will change for the same controller from time to time. It can be caused by hardware revisions and such. Some controllers have the same GUID across platforms, some don't, which means you can't just copy-paste the mappings (unfortunately). Anyways, no harm done. I'll wait for a PR update or a new PR. Cheers.

r-lind:
Forgive me for assuming it could, but since it remained the same on all Logitech Dual Action entries across the different platforms with different versions on different platforms I come to no other conclusion. And I only have a v2.0 to test with, but since the entries have no other difference except OS, I would personally be confident to find all the combinations likely to turn up even though untested.

p-groarke:
Would you rather I add the mapping you pasted in the issues myself? The problem with that is you wont get credit and wont get listed in collaborators page.

If not, I recommend restarting from scratch. Create a new branch or delete/re-fork the project and just add your new mapping. The PR should pass tests.

PR #190, opened 2018-01-27
Fix guide button mismapping for Hori Mini 4

jdfeng:
Sorry! I realized today I'd mis-mapped the guide button for the Hori Mini 4 controller (I used the "touch" button instead of the PS4 button.  Although the touch button's a lot more comfortable, it probably would confuse others.)


p-groarke:
Sorry for the long delay, RL happened :) Thanks for sending the fix!

PR #191, opened 2018-01-31
Update 8bitdo controlers on Linux

Darknior:
- Replace the two 8bitdo SFC30 where there is not the DPAD Keys
- Add 8bitdo SF30 Pro & NES30 GamePad

All generated with the SDL2.0.7 controllermap on my Linux PI3.


Darknior:
I think all is fine now, but your bot can't understand that i make 3 commit from GitHub.

p-groarke:
I'm sorry to tell you all is not fine. You can read the details of the unit tests for more information :
```
Cannot modify upstream SDL header mapping at line #233
If you have problems with an official SDL mapping, please report the issue or send a pull request to the SDL project : libsdl.org.
```

This database is a super-set of the SDL internal database. You will need to send your PRs to them, and if your mappings are accepted we will get them at the next SDL version.

Other potential issues : You are removing the axis directions. The 8bitdo reports its dpad as an thumbstick, that is why it is mapped to a joystick. Real dpads would be hats like h0.X.

Good luck with your SDL pull requests! Thx anyways for the PRs.

Darknior:
Ok thanks i will open some issues...

> Other potential issues : You are removing the axis directions. The 8bitdo reports its dpad as an thumbstick, that is why it is mapped to a joystick. Real dpads would be hats like h0.X.

I understand what you write, i know it, but it is not a real thumbstick too ... it is not really possible to use it like a real thumbstick ... we use them like a normal cross to play old games. 
You really think i must configure them like Thumbstick to open my issues ?

Thanks

p-groarke:
Hey, sorry I missed your response. I'm not sure, but I'd lean towards a thumbstick if it returns axises (for dpad). Your best bet might be to get an external xinput tool. The db isn't really meant for remapping in a "more comfortable" way.

All the best!

PR #194, opened 2018-02-04
added mapping for MP-8866 Super Dual Box PS2 adapter

Dregu:
This thing is centuries old, why is it missing :o


p-groarke:
Your windows mapping doesn't have the back button, why is it missing :o

p-groarke:
Thank you for the PR :)

PR #195, opened 2018-02-19
add Street Fighter IV Fightstick TE (Linux)

mbenkmann:
add Street Fighter IV Fightstick TE (Linux)
Note: The first commit was a reorder that check.py --format performed on the unmodified file. I guess the previous change was done without running check.py --format or there are version differences in check.py that sort differently.


p-groarke:
Yes that happens with mappings that have the same name once in a while (its not really deterministic that way).

Thank you for the mapping!

PR #196, opened 2018-02-19
Make --format sort order deterministic

mbenkmann:


p-groarke:
Looks good. Out of curiosity, why are you using a null char in between the name and id?

p-groarke:
I personally prefer this way of concatenating strings:
```
("%s%s" % (x[1].name, x[0])).lower())
```
I don't mind adding it myself, but I'd rather you get the credit in commits.

mbenkmann:
The null char is so that shorter names are sorted before longer ones. E.g. "Controller Foo" is sorted before "Controller Foo v2". Without the null, the sort order would depend on the GUID.

With respect to the commit. Just merge my code and then rewrite it as you see fit. But I would recommend including the null char.

PR #203, opened 2018-04-08
NEXT Classic Controller Modified to support same ID controllers.

Superwaitsum:
Hello, I added compatibility for the SNES and Sega controllers that have the same ID that the NES controller used in 204 and 205 old database,

Modified A button (b0 -> b2) of the NES controller to give more compatibility to the NES and Sega controllers.
The Sega controller doesn't have the Z->RB (recommended in the readme) due to the SNES having L and R working like in the main gamecontrollerdb (which also miss Sega controller, but at least has SNES) .

Also modified linux platform, since it should work the same.

Checked the files with the check.py.

Sorry if the description looks awkward, it's my first PR.


p-groarke:
Hey no problem! It's good to see some fixes to the legacy dbs as well :) Thank you

PR #208, opened 2018-05-27
Add GPD Win 2 Controller (Linux)

flibitijibibo:
Not much to see here; this is basically a clone of the 360 controller but with a different GUID.


p-groarke:
Thank you for the pr!

PR #209, opened 2018-06-05
Added another GUID for the 8Bitdo Zero for Mac

nathanhoad:
I have an 8Bitdo Zero but its ID was missing from the db.


p-groarke:
Ty!

PR #215, opened 2018-07-17
Update gamecontrollersdb.txt

TyanColte:
Added XBox One elite wired controller.


p-groarke:
Can you format the entry and make sure it is only 1 commit. Thank you.
`/check --format gamecontrollerdb.txt`

p-groarke:
Looking good, thank you!

PR #219, opened 2018-09-12
DirectInput Xbox 360 controller updates

Jimbly:
Updated wired mapping to include right trigger
Add wireless mapping

Notes: it looks like the wired mapping was never updated with the new mapping scheme (allowing +a2 and -a2 for triggers).
These mappings are only used in games that call `SDL_SetHint(SDL_HINT_XINPUT_ENABLED, "0");` to read all controllers as DInput (which is required if you want to support more than 4 XInput-capable controllers).
This fixes an issue in [Splody](https://store.steampowered.com/app/467810) not recognizing triggers reasonably.


p-groarke:
Thank you!

Jimbly:
Glad to help, thanks for merging!

PR #222, opened 2018-09-27
Add other X-Box One S Controller for Linux

akien-mga:
Note: I haven't tested it myself, it's based on findings from @MoustafaC and @aaforcebox on godotengine/godot#18167.


p-groarke:
Thx for the PR. If it makes it upstream, I'll import it from there. Shouldn't require any more work from you. At that point, if you want to modify the mapping, it'll have to come from upstream (as we don't touch mappings from SDL project). Cheers

stevoisiak:
To be specific, this patch is for the [Model 1708](https://www.ifixit.com/Device/Xbox_One_Wireless_Controller_Model_1708), also known as the Xbox One S controller.

mstfacmly:
It bears mentioning that if Godot pulls from upstream SDL for the inputs, then this mapping would be unnecessary.

stevoisiak:
@mstfacmly I don't believe this mapping is currently part of SDL.

stevoisiak:
This mapping has been added upstream to SDL. ([Bug](https://bugzilla.libsdl.org/show_bug.cgi?id=4325), [patch](https://hg.libsdl.org/SDL/rev/5dc13016cf34))

PR #224, opened 2018-10-15
Add Switch Joy-Cons

porglezomp:
This adds Nintendo Switch Joy-Con support as "Joy-Con (L)" and "Joy-Con (R)". These are for the Joy-Cons in their horizontal position, so they only have the x/y/a/b buttons, the left stick (including press), the shoulder buttons, and the back/start buttons.
The Joy-Con's vertical orientation trigger and button aren't mapped to anything.
This maps the back button to whichever of the +/-/screenshot/home button is on the left, and the start button to whichever is on the right.

The mappings were created / tested on macOS, so I'm not sure if the other platform mappings are correct. All of the Switch Pro Controller mappings are the same across all platforms though, so I just duplicated my mappings between platforms.


porglezomp:
Ah, I just noticed that there is a single mapping for Joy-Con (L) on macOS already in the db. It disagrees with mine on whether there should be separate start/back mappings.

It uses the - button as a start button, whereas I make that the back button, and the screenshot button the start button. If the intended purpose is to map the layout onto an xbox controller, then I *think* my approach is better for that, but I'll accept feedback.

porglezomp:
What my layout does, for reference:

joycon

p-groarke:
Thanks for the PR. I disagree using direction like that. + or - should be start, and "home" or "screenshot" should be back (select). Other than that everything looks good to me.

porglezomp:
Okay, that makes sense, I've swapped it.

p-groarke:
Perfect

PR #231, opened 2018-11-16
Add Logitech RumblePad 2 USB for Mac OS X

gcatlin:
I inverted `lefty` and `righty` in order to have positive values when the stick is pushed forward and negative values when the stick is pulled back. I'm also assuming `leftx` and `rightx` should be negative when pushed left and positive when pushed right. Lemme know if I made bad assumptions.

Thanks for creating and maintaining this project!


p-groarke:
Thx for contributing!

PR #234, opened 2018-11-22
Microsoft X-Box One pad (Firmware 2015)

MCMrARM:
Add a mapping for Microsoft X-Box One pad (Firmware 2015)

Generated using SDL2 2.0.9's controllermap.


p-groarke:
Thank you

PR #242, opened 2019-01-16
Updated and added HORIPAD mini4 config for MAC and Linux

jlahman:
Added config for HORIPAD mini4 on mac. Updated name of HORIPAD mini4 on linux from HORIPAD mini4 (PS4), I previously thought that was convention but realized it isn't. Fixed guide button binding for the mini4 on linux.

I ran check.py --format, that is why an unrelated config has been updated.


p-groarke:
ty!

PR #249, opened 2019-03-08
Add alternate Joy-Con, alternate SFC30, Saitek P220 mappings

AliceLR:
Here are some mappings found by @insidious611 and KKairos while testing gamecontroller integration for AliceLR/megazeux#157. Included are:

* Added alternate Joy-Con mappings for Windows (different version number).
* Added alternate Joy-Con mappings for Linux (bluetooth bus type instead of USB).
* Added alternate 8Bitdo SFC30 mapping for Windows (different VID and PID).
* Added Saitek P220 Windows mapping. This controller is a little weird; it has shoulder buttons and six face buttons but no start/back equivalents. Instead, it has a "throttle" axis controlled by two buttons that doesn't really map well to anything. We put start/back on the extra face buttons (the Sega recommendations don't really make sense here) and left the "throttle" unmapped.


p-groarke:
Thx!

PR #252, opened 2019-03-17
XBox One controllers. Wired + wireless when SDL_XINPUT_ENABLED=0 for platform:Windows

supersmo:
Added controller mappings for XBox One controllers for use with SDL_XINPUT_ENABLED=0.
When SDL_XINPUT_ENABLED=0 the triggers are on the same axis and the guide button cannot be accessed.

Was a bit unsure of the controller naming but in the end I went with what was displayed in the control pannel and added " -Wireless" and " -Wired".

Both mappings were generated with an Microsoft XBox One S controller.


p-groarke:
thx!

supersmo:
Wow that was fast! 
And thanx for maintaining this controller mapping db!

PR #253, opened 2019-03-30
Added another ShanWan PS3/PC Wired GamePad (DEXP G-6) for Windows and…

xoreaxebx:
… Linux


p-groarke:
Could you name the XInput version more descriptively? ty

xoreaxebx:
I dunno how it named indeed. Steam names this so and I haven't seen in the guideline that I may rename. Maybe you need mention about this in readme because there are lot of "USB gamepad" and similar already. 

Renamed btw

p-groarke:
Naming Convention

    If the controller has a unique name, use that (ex. XBox Controller, XBox One Controller).
    If the controller doesn't have a unique name, or when in doubt, use brand + model (ex. ACME GA09, 8Bitdo NES30 Pro).
    You can add Wireless for controllers with such an option. (ex. XBox Wireless Controller).

;)

All good though.

PR #254, opened 2019-04-02
Add PowerA 1428124-01 for Linux

isnt-alice:


p-groarke:
thx, somethings up with travis, second run was good 👍

PR #256, opened 2019-04-11
Add Mayflash GameCube Adapter and Jess Tech GGE909 Recoil for Linux

LesserHydra:
For the GameCube adapter, A/B/X/Y are mapped as named since the controller has a nonstandard layout. Hopefully this is in line with the mappings for other platforms, though I can't test those to be sure.

The Recoil, on the other hand, is pretty standard and just lacks a 'guide' button.


offalynne:
standards clarification for gamecube adapters and saturn style pads (6 face, 2 shoulder buttons) would be really helpful as they’re currently inconsistent between multiple mappings in the db. i personally think xi:gc should take the format a:a b:x x:b y:y to keep north/south/east/west face buttons approximately equivelant, in any case thanks for clairfying how you mapped it!

p-groarke:
Thx for the PR. You make a good point about the gamecube controller. The correct mapping would be what @offalynne mentions (a:a b:x x:b y:y). I will add it to our readme soon.

I do have to ask you to update your PR with the mapping change :/

@offalynne The sega saturn mapping guide is already in the readme ;)

offalynne:
pardon the digression but saturn controllers have shoulder buttons! is the implication that the right shoulder should be mapped to LT then ?

p-groarke:
@offalynne Yep. I guess I can add the shoulders when I update the graphic for the gamecube controller.

LesserHydra:
No problem, that makes sense. Thanks both for the clarification

p-groarke:
thx!

p-groarke:
@LesserHydra @offalynne The image has been updated. Sorry for the delay, I lost the original vector file.

PR #257, opened 2019-04-12
Add 8Bitdo FC30 Pro for Linux

xordspar0:
Aside from the appearance, the 8Bitdo FC30 Pro is identical to the NES30, but it has a different name and GUID.


p-groarke:
ty!

PR #259, opened 2019-04-30
Update gamecontrollerdb.txt

XenoAmess:
another type of betop.


XenoAmess:
wait,something is not so right here due to some mis-operations.
I will close it and try another time.

PR #261, opened 2019-05-01
update gamecontrollerdb.txt

XenoAmess:
1. add another betop controller
2. sorry for the misoperation I made before.


XenoAmess:
something not right here...I will see what happened

PR #263, opened 2019-05-14
Add Hori Pokken Tournament DX Pro Pad gamepad

joow:
Add mappings for Hori Pokken Tournament DX Pro Pad gamepad for Windows, macOS and Linux as well.


p-groarke:
Thank you!

PR #264, opened 2019-06-13
Add Gravis Eliminator GamePad Pro (Windows, Mac, Linux)

AliceLR:
Adds Windows, Mac, and Linux mappings for the Gravis Eliminator GamePad Pro controller. Tested with Windows 10, Snow Leopard, and Linux Mint.

This controller is mostly standard aside from the "dpad", which is actually a large and hypersensitive analog stick ([picture](https://i.imgur.com/s0JB1X9.png)). It's hard enough to use that there is a "Precision" button in the middle (not a guide button) that can toggle its sensitivity, not that it helps much.


p-groarke:
Ty Alice

PR #266, opened 2019-06-22
Fixed issue with PS3 controller where left trigger stays pressed

Pakoach:
Here is the PR that solves the PS3 controller issue.


p-groarke:
Thx for the PR. Unfortunately, this is an upstream mapping (it comes from SDL itself). We don't modify those. If you want to fix the issue, you can submit a ticket and/or pull request to the SDL project directly. Once they merge it and release a new version, we will get the fix. Good day.

PR #267, opened 2019-06-23
Added PDP Wired Controller for Xbox One (Linux)

CookiePLMonster:
This PR adds support for PDP Wired Controller for X1:
https://www.pdp.com/en/pdp-wired-controller-048-082-na-wh01

I created this mapping via Steam Link app on Raspberry Pi - there the controller manifested itself as "XInput Controller", so I renamed it to something more accurate.


p-groarke:
Ty!

PR #270, opened 2019-06-30
Snes retroport and gamestop branded third party 360 controller

MichaelBuckley:
Hello,

Thank you for mainaining this database of USB controllers. I have a couple of controllers I use at home that would like to add so more games can support them. One of them is an SNES to USB adapter, and is already in the database for Windows and Linux. This pull request adds an entry for Mac.

The second controller is a GameStop-branded third-party XBox 360 wired USB controller that I somehow inherited. This pull request adds it to Mac and Windows, but I was unable to get it working on Linux.


p-groarke:
Nice! For the windows and linux xbox controllers, can you rename them to "GameStop Xbox 360 Wired Controller". I think that would be more clear. Ty.

MichaelBuckley:
Sure thing. I've updated the pull request.

p-groarke:
👍

PR #271, opened 2019-07-01
Xbox Adaptive Controller for MacOS & Windows

aarnoud:
This adds support for Xbox Adaptive Controller for MacOS & Windows

https://www.xbox.com/en-US/xbox-one/accessories/controllers/xbox-adaptive-controller


offalynne:
great work. would be *excellent* if you were able to make a PR with this to the upstream SDL db; seems like they prioritize first-party devices.

p-groarke:
Nice! thx

PR #273, opened 2019-07-14
Add Afterglow Xbox 360 controller for Linux

alexarice:
String generated with steam big picture, not sure if the format is correct.
This looks similar to some of the other controllers there but this was not getting picked up correctly when I tried to run openmw (even after replacing their list with the most up to date one)


alexarice:
Check seems to have failed referencing an upstream repository, should I be sending prs there instead?

offalynne:
It's already in the db from upstream: https://github.com/gabomdq/SDL_GameControllerDB/pull/273/files#diff-cc38bbf82d721173b43f2c8885a37872R383

You need to check before submitting a PR: 
https://github.com/gabomdq/SDL_GameControllerDB#check-your-mappings

alexarice:
Thanks for the links, I'll take my issue upstream

PR #274, opened 2019-07-19
Improve mapping for Xbox Adaptive Controller on Mac OS X

grum:
When using https://github.com/360Controller/360Controller on Mac OS the controller (wired) only reports 6 axes, 15 buttons and 0 hats. The old mapping seems to be a copy `030000005e040000fd02000003090000` (Xbox Wireless Controller) right above it and didn't function for me since it refers to button 16.

Maybe @Aaarnoud can test it as well?


offalynne:
The upcoming MacOS Catalina has native Xbox One device support so it will be interesting to see how that impacts SDL map results. I suspect it will end up motivating a gradual move away from the presently popular existing userland drivers on MacOS.

p-groarke:
thx!

offalynne:
if the map @Aaarnoud provided was functional this is actually a regression because it relies on a third-party driver which is very likely not supported in upcoming MacOS :(

PR #275, opened 2019-08-02
Adding PDP Afterglow Prismatic 048-007-NA (Linux)

One-Nub:
This seems to be an older model versus what actually is on the market now (newest is 048-121-NA), bindings are the same as the other Afterglow controllers though


p-groarke:
ty!

PR #277, opened 2019-10-01
Added Xbox One S controller over bluetooth

mrhelmut:
Hey there!

It seems that the latest Windows 10 versions messed around with Xbox One S controllers IDs when connected over bluetooth. It strangely adds ```7801``` at the end of the ID.

So here's a mapping to make Xbox One S controllers to work over bluetooth with the very latest Windows 10.

Thanks for the cool work being done over here.

Cheers!


offalynne:
Good stuff, would you consider pushing this upstream to SDL? They prioritize support for first party devices like this. Theoretically the mapping shouldn’t be necessary on Windows  as the gamepad library should be deferring to a generic XInput map, but I’ve noticed this happens on Win8.1 over BT as well.

 SDL’s repo is here https://hg.libsdl.org/SDL

mrhelmut:
Will do whenever I can contribute, for sure.

p-groarke:
@mrhelmut Thx for the PR. Feel free to reopen this if SDL don't want it ;)

mrhelmut:
I believe it has been fixed upstream. I've updated our game to 2.0.10 without the mapping and it seems to work.
Weird thing is that we were running on 2.0.7, and Gamepad Tool (also using 2.0.7) found a default mapping, while ours didn't. I believe we were running on a pre-release of 2.0.7 and it got fixed.

p-groarke:
Perfect. Didn't see 2.0.10 was out, will rebase this weekend.

PR #279, opened 2019-10-07
adding 8BitDo maps and Mayflash Magic NS

offalynne:
thanks to @heavyviper and Zhang Siho of 8BitDo for assistance


p-groarke:
Ty! :)

PR #282, opened 2019-10-15
Fix for G-Shark GS-GP702 Controller righty Analog

lordloki:
´righty´ value is a3, not a4 (tested using the controller).


offalynne:
This map comes from upstream via SDL2 and so can not be modified in this DB, hence the CI check failure. The GUID in question is from a generic board shared with over a dozen devices with conflicting mappings, mostly of Chinese manufacture; it really should be removed from the database altogether since there’s no demonstrable candidate with best market share to my knowledge. Out of curiosity, do you know the model/make/number/name corresponding to your device in particular? An lsusb dump via Linux (or cygwin) would also be handy for future posterity, if you’re able to provide one.

lordloki:
It was reported from @UnidayStudio on our repository (https://github.com/UPBGE/blender). 
@UnidayStudio could you supply more info about your controller?

Thanks

p-groarke:
Closing since we don't modify upstream mappings. Feel free to continue the discussion.

PR #283, opened 2019-11-15
added GameSir GameSir-T2a mapping

schnere:


p-groarke:
Hi, you mapping has issues we've seen in older versions of the SDL mapping tool. Are you running the latest version? Can you update your mapping tool and rerun the mapping? Thx.

PR #286, opened 2019-11-24
Add Xbox Elite Series 2 Wireless

duganchen:
My test system is Fedora 31, stock kernel and drivers, with the controller running the latest firmware. This works well with the tests I did. The button maps are correct in GamePadTool, and they're correct in Dead Cells based on the few minutes I playtested with it.


duganchen:
Sorry about this. I'm going to submit a new request with a non-messed-up history.

PR #288, opened 2019-12-03
Support for SteelSeries Stratus Duo on Linux

Lars-Saetaberget:
This controller has different GUIDs depending on whether it's run in cable or dongle mode. Hence the two entries.


p-groarke:
Thank you

PR #291, opened 2019-12-04
Add Google Stadia Controller mappings for Mac and Linux

aaronfranke:
These were all generated by Steam, with the controller plugged in with a cable. The Mac mapping was generated using a macOS Mojave virtual machine with USB passthrough.

I re-did the Windows mapping to be consistent with the Mac and Linux ones, but I checked that the mapping is the same as before and it is.

Also, I noticed that the README had some incorrect information about the values that Steam outputs, so I updated the README with instructions that were relevant to my experience. Steam does output `platform:` but at the beginning instead of the end, and for some reason the controller name started with `dev:gen:` (this happened on Windows, Mac, and Linux).

Also, I have an Android phone and a USB-OTG cable, but what's the easiest way to make an Android mapping? Do I really have to make an app using the SDL2 source code? It would be nice if there was an app on the Google Play store for configuring the controller.


offalynne:
they're already upstream, see https://hg.libsdl.org/SDL/rev/f337f1e13756

> Do I really have to make an app using the SDL2 source code?

yes

p-groarke:
Thanks @offalynne :)
@aaronfranke If you wish, you can create a seperate PR with just the readme update. I wouldn't  change the example mapping. You can simply add that if the controller name may need some cleaning up, for ex if it contains dev:gen: Thx

p-groarke:
As far as I know, you can use GLFW and love engine, which supports parsing the db. Unreal uses some mappings internally, though I don't know if it can parse the file. I believe Unity uses them underneath the hood as well, but again I'm not sure if you can load the db with their APIs.

aaronfranke:
@p-groarke The point of changing the example is because the old example doesn't feature the things I mentioned in the README (neither `dev:gen:` nor `platform:`).

So, are you going to add these from upstream then? As far as I know, redundant bindings are looked favorably upon, and they are necessary for projects that don't use SDL2 (like Godot).

p-groarke:
What I meant is that `dev:gen:` seems specific to your controller. We've never heard of any other mapping with it, so just mentioning to double check the name seems enough. Feel free to reopen this PR with the changes if you prefer that BTW,

Yes upstream mappings get pulled at every official release of SDL. Our DB contains both, the official mappings and the community editions. We never modify official mappings.

aaronfranke:
@p-groarke Ok, done, it's just README changes and nothing changed in gamecontrollerdb.txt. I did add `platform:Windows,` to the example because Steam does indeed output the platform now, so this is necessary to keep the example relevant.

Note that my changes don't show up until the pull request is opened (GitHub feature, closed PRs don't display pushes for some reason).

p-groarke:
Weird, I can't reopen the PR. Can you? You can just open a new one. Sorry for the extra steps.

aaronfranke:
Really? That's really weird. I can't open it, when maintainers close it non-maintainers can't open it. I am very surprised that you can't re-open it, it's probably a glitch or something? https://github.com/gabomdq/SDL_GameControllerDB/pull/293

Just a tip, maybe give feedback before closing immediately in the first place :)

p-groarke:
Closing PRs has never been an issue in this project.

Also, please stop necroing old PRs and issues.

Issue #289, opened 2019-12-04
NEXT SNES Controller right shoulder mapping is wrong


leiradel:
This controller's right shoulder is mapped to button 6, but the corresponding `SDL_Joystick` reports button 5 being pressed when I press the right shoulder.

Changing the mapping to `rightshoulder:b5` solves the issue.

offalynne:
Which GUID?

leiradel:
Sorry, it's `030000001008000001e5000010010000`.

p-groarke:
Please open a PR with the fix, thank you.

PR #294, opened 2019-12-05
Map the correct button for the right shoulder

leiradel:
As requested in https://github.com/gabomdq/SDL_GameControllerDB/issues/289


p-groarke:
Thanks for opening the PR! It seems this mapping comes from the official SDL database. You'll have to open up your PR at the official SDL project, since we don't modify those. Your changes would just get erased at the next upstream update.

Let me know if you have any questions.

leiradel:
Ok thanks

PR #295, opened 2019-12-05
Add some controllers to Linux controller database.

lightningterror:
HORI CO. LTD. HORIPAD S
Logitech F510 Gamepad
Saitek PS2700 Rumble Pad
SteelSeries Stratus XL


p-groarke:
Thanks a lot for all the mappings!

PR #302, opened 2019-12-30
Add support for 8BitDo SN30 Pro on Mac OS X.

wtetzner:
Add support for the 8BitDo SN30 pro controller on Mac OS X.

8BitDo SN30 pro


offalynne:
good stuff

PR #314, opened 2020-02-12
Added vJoy virtual controller and Raphnet GC/N64 to USB GCN mappings

tdeeb:
This PR adds mappings for the vJoy virtual controller along with Windows and Linux mappings for a GCN controller with the Raphnet GC/N64 to USB adapter.


p-groarke:
ty!

PR #316, opened 2020-02-13
Fixed one axis for 8BitDo SN30 Pro

FrodeSolheim:
Verified with latest firmware for the controller


p-groarke:
@offalynne Would you like to double check this one? I can't remember if it was part of your large PR with 8bitdo controllers.

FrodeSolheim:
Always a good idea to doublecheck :) For reference, I created my config with 8BitDo SN30 Pro firmware v1.33 on macOS 10.14.6.

offalynne:
looks good to me

p-groarke:
thx!

PR #320, opened 2020-02-16
Fix 8BitDo SN30 mappings for Windows

FrodeSolheim:
Not sure what's going on with the existing mappings in the database (The SN30 only has a d-pad, a/b/x/y buttons, left/right shoulders and select/start buttons). I've created new ones from my 8BitDo SN30 with latest firmware (USB + Bluetooth mode).

(The existing 8BitDo SN30 GP entry matches my 8BitDo SN30 in Dinput mode over Bluetooth. The GP is just a different color scheme on the controller itself, so it makes sense to just call it 8BitDo SN30 since probably all color variants have the same hardware and GUID).


offalynne:
> existing mappings
this was my mistake, thanks for the corrections

FrodeSolheim:
Thanks for confirming :)

PR #321, opened 2020-02-16
Axes fixed for 8BitDo N30 Pro 2 / Windows

FrodeSolheim:
Mappings created with latest firmware version (v6.21) for 8BitDo N30 Pro 2 and SDL 2.0.10.

The axes are weird on Windows: leftx is a0, lefty is both a1 and a2 (same value on both axes),
rightx is a3, and righty is both a4 and a5 (same value on both axes).

I thought this was a bit weird, so double-checked with Steam gamepad configuration, and got the same result.

(I get some strange results on Linux and macOS as well, but there, dpad left/right triggers the same axes (in addition to dpad movement) as left/right triggers (in addition to buttons for the triggers. So on all platforms, the device behaves a bit strange but not equally so. Could be buggy firmware causing undefined behavior or something)


FrodeSolheim:
I was a bit interested in figuring out the strange behavior (since the axis duplication didn't happen on Linux and macOS). I found the problem, and explained it here https://discourse.libsdl.org/t/duplicated-joystick-axes-with-sdl-dinputjoystick-and-8bitdo-n30-pro-2-with-suggested-fix/27247 along with a suggested fix for SDL, in case you're interested.

(The "strange behavior" on Linux and macOS where the dpad left/right and triggers shares axes are actually a firmware problem not OS/SDL, I debugged the HID reports being sent from the device).

PR #322, opened 2020-02-19
Add Padix Co. Ltd. Rockfire PSX/USB Bridge

Ziemas:
An old Playstation to USB adapter.


lightningterror:
@gabomdq I noticed this controller/adapter is no longer in the DB.

PR #343, opened 2020-05-12
Fixes right stick axes for SF30 Pro (discontinued) on Windows. 

matthewlim:
- Previously, any direction on the right stick would register as axis 2. After this change, the X and Y axes on the right stick register as controller axes 2 and 3, respectively.


matthewlim:
Yes - the firmware is V1.33: "2019-11-29 SN30pro+SF30pro V1.33"

offalynne:
Great. If you’re able you might want to take these to the SDL HG repo, quite a bit has changed upstream since the whole DCMA debacle with this community repo and it’s fallen out of parity with the db in the SDL source which currently has the outdated maps for these GUIDs.

matthewlim:
no kidding. is that what everyone normally does? merge here and then into the mercurial repo for SDL? 

for the purposes of this repo, is there anything else i need to do to have it merged?

also, what happened with DMCA? 

thanks!

offalynne:
>  is that what everyone normally does?

no, but i think it's appropriate in this case for a few reasons

> for the purposes of this repo, is there anything else i need to do to have it merged?

no, the maintainer will likely merge when he gets around to it

> what happened with DMCA?

a former maintainer claimed against the repo for several contributions including a sorting script that handled, amoung other things, keeping this db in parity with upstream changes that have subsequently been purged

matthewlim:
understood - i will try to follow up with the SDL repo. Thank you for the explanation, very much appreciated. stay safe!

PR #348, opened 2020-05-24
Add Mayflash WiiU alternate GUID

hoontee:
Same mappings, different GUID.


hoontee:
Fixes godotengine/godot/issues/38869.

akien-mga:
For context, the mapping for this GUID used to be provided in `gamecontrollerdb_204.txt`. I don't know if it was left out from `gamecontrollerdb_205.txt` and then the current `gamecontrollerdb.txt` on purpose or by mistake.

offalynne:
This is the old GUID format and those maps are no longer in the repo. If you have access to the hardware, consider contributing an updated GUID for the map. If you want to use  maps from the older DB versions yourself SDL will support them, but they’re deliberately omitted from this DB as the new format contains additional disambiguating data.

akien-mga:
Thanks, this PR should likely be closed then, it seems like we need to review the way we retrieve GUIDs in Godot.

PR #363, opened 2020-07-13
Fix 4c05 6802 mapping.

madmiraal:
During the update to SDL 2.0.9 (dd6640d10) the mapping for 030000004c0500006802000000000000 was significantly changed. This PR reverts that change.

2.0.8 Mapping |   | 2.0.9 mapping |  
-- | -- | -- | --
leftx | a0 | leftx | a0
lefty | a1 | lefty | a1
rightx | a2 | rightx | a2
righty | a3 | lefttrigger | a3~
  |   | righttrigger | a4~
  |   | righty | a5
back | b0 | y | b0
leftstick | b1 | b | b1
rightstick | b2 | a | b2
start | b3 | x | b3
dpup | b4 |   |  
dpright | b5 |   |  
dpdown | b6 | leftshoulder | b6
dpleft | b7 | rightshoulder | b7
lefttrigger | b8 | start | b8
righttrigger | b9 | back | b9
leftshoulder | b10 | leftstick | b10
rightshoulder | b11 | rightstick | b11
y | b12 | guide | b12
b | b13 |   |  
a | b14 |   |  
x | b15 |   |  
guide | b16 |   |  
  |   | dpup | h0.1
  |   | dpright | h0.2
  |   | dpdown | h0.4
  |   | dpleft | h0.8

Looking at the [SDL change logs](https://hg.libsdl.org/SDL/log), this change was made with [changeset 12011: Added support for the PS3 controller driver in PlayStation Now](https://hg.libsdl.org/SDL/rev/3eab73f3e17e). This is a substantial change to a mapping for a change that was "adding support". Therefore, I suspect this change was made in error; especially since the other mappings touched by this change weren't significantly changed: 2509 0500 only had its name changed, and the other 4c05 6802 entry (for Linux) only had the order of the entries changed. Furthermore, 4c05 6802 was given the mapping of 2509 0500 that had its name changed to be the same i.e. this could have been a cut and paste error.

This [issue](https://github.com/godotengine/godot/issues/39992) was picked up by a Godot user when Godot applied the new mappings.


akien-mga:
> Furthermore, 4c05 6802 was given the mapping of 2509 0500 that had its name changed to be the same i.e. this could have been a cut and paste error.

It doesn't seem to be the exact same mapping though, the first buttons match but then the similarity ends:
```
"03000000250900000500000000000000,PS3 Controller,a:b2,b:b1,back:b9,dpdown:h0.8,dpleft:h0.4,dpright:h0.2,dpup:h0.1,guide:,leftshoulder:b6,leftstick:b10,lefttrigger:b4,leftx:a0,lefty:a1,rightshoulder:b7,rightstick:b11,righttrigger:b5,rightx:a2,righty:a3,start:b8,x:b0,y:b3,",
"030000004c0500006802000000000000,PS3 Controller,a:b2,b:b1,back:b9,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b12,leftshoulder:b6,leftstick:b10,lefttrigger:a3~,leftx:a0,lefty:a1,rightshoulder:b7,rightstick:b11,righttrigger:a4~,rightx:a2,righty:a5,start:b8,x:b3,y:b0,",
```

I guess that if this revert is correct, it should also be changed in SDL: https://hg.libsdl.org/SDL/file/465afae5eb7e/src/joystick/SDL_gamecontrollerdb.h#l232

offalynne:
that VID/PID (054c 0268) matches the DualShock. there are lots of knockoff devices that report similar, see the equivalent Linux mappings. you should prompt your user to install the Playstation Now driver.

akien-mga:
> in general Godot should probably defer to the upstream mappings at this point, since this repo is no longer kept in parity with SDL.

Godot doesn't use SDL, but implements the same mapping format, and uses `gamecontrollerdb.txt` from this repo to populate our mapping DB.

Do you recommend that we parse https://hg.libsdl.org/SDL/file/830a4ce7456e/src/joystick/SDL_gamecontrollerdb.h instead?

If so, it's not clear to me what is the use case for this SDL_GameControllerDB repository?

offalynne:
>Do you recommend that we parse [...] instead?
At present, yes.

This DB is as you describe, but currently it’s out of parity with the upstream DB which by it’s nature (being shepherded by the SDL developers) will be be more reliable.

It’s long been the case that upstream GUIDs aren’t modified in this DB, in any case.

gabomdq:
This DB is a superset of what SDL includes. Originally Sam and Ryan from SDL didn't want to add too many mappings to SDLs that they couldn't verify, so this DB was created as an unofficial repository for these mappings. Now it seems they are adding things faster, however for games already released updating the built in database requires waiting for a new release (or building your own from hg), replacing the SDL binary (and in some cases rebuilding the game), so this database is an easier path to update in that case...which is why you will see duplicates from "upstream".

offalynne:
Regardless, this PR shouldn't have been merged... it's the wrong mapping for DualShock 3 controllers and breaks operation of the real thing in deference to a marginal aftermarket product.

PR #368, opened 2020-07-20
update from SDL

offalynne:
Update mappings based on upstream SDL's built in controller database.


gabomdq:
There are a few removals here, what's the reason for that?

PR #376, opened 2020-08-09
fix dpad mapping of Razor Onza Tournament Edition on Linux

klianc09:
For me the dpad wasn't working on linux. The dpad was mapped to buttons but it actually is a hat switch (for me).

I have absolutely no idea if mappings across linux versions are always the same, or whether this could break the mapping for someone else. So I cannot tell if #18 initially created a faulty mapping, or it actually worked for them. Or maybe it worked back then, but it's different now due to some update, I have absolutely no idea how this works. 

I would be very grateful, if somebody could tell me if I need to take care of something when creating pull requests for Linux mappings.

If this fix however is correct, I'm very suspicious that the mapping for `Razer Onza Classic Edition` has the same error, since that is the only other Razer mapping that uses buttons for the dpad, all other Razer mappings use the hat switch. However, I have no Razer Onza Classic Edition to actually verify that.


offalynne:
Probably a driver update and/or different udev rules. Are you using xboxdrv? You could compare with/without perhaps. The hat mapping is probably closer to "correct", but it's a little hard to determine which map should be more generally accepted.

klianc09:
I was not using xboxdrv. So I installed xboxdrv for testing, and with xboxdrv BOTH mappings work ... strangely enough.

Unless I screwed something up majorly, I think it's safe to say my mapping should be accepted, since it works with and without xboxdrv and the current mapping only works with xboxdrv.

PR #385, opened 2020-09-05
Add pdobq adapter under Windows

sorenbug:


sorenbug:
Don't merge this yet, left joystick is broken

sorenbug:
The joysticks and triggers will require calibration under Windows' built-in "Calibrate USB Joysticks" tool to count as directional inputs as they do not naturally exceed 0.8 on one half of the axis. Once that's done, the controller will work.

offalynne:
Are you sure the right stick axes are inverted @sorenbug ? They aren't in RetroArch https://github.com/libretro/retroarch-joypad-autoconfig/blob/master/xinput/MAYFLASH_GameCube_Controller_Adapter.cfg

sorenbug:
@offalynne The pdobq adapter inverts the right stick for some odd reason. None of my friends with different brands of adapter have this issue, it seems to be exclusively a pdobq issue.

offalynne:
Hm, tricky. Better to remove the inversion modifiers on the Windows map in my opinion since a lot of other GC > USB devices match this GUID and afaict none of them have these axes inverted, though otherwise the match maps. Less an issue on Linux since there’s additional data in that GUID that (hopefully) disambiguates this device in particular.

offalynne:
Also just for posterity I want to make note that this contributor has chosen to match the XInput face buttons to those on the GC pad — B is “b” and X is “x”, which is not necessarily wrong (there seems to be some ambiguity regarding policy on this kind of thing in the upstream SDL maps lately) but it _used to_ be (loose, unenforced) policy in this repo that we swapped B and X on GC pads to keep face buttons in the same approximate position ignoring button label inconsistencies. No idea if that’s actually reflected in other mappings, and probably not worth worrying over.

sorenbug:
> Hm, tricky. Better to remove the inversion modifiers on the Windows map in my opinion since a lot of other GC > USB devices match this GUID and afaict none of them have these axes inverted, though otherwise the mappings match. Less an issue on Linux since there’s additional data in that GUID that (hopefully) disambiguates this device in particular.

Really? There's other devices that match this GUID? _sigh_ If only Windows had a built-in option to flip joysticks 😕

PR #387, opened 2020-10-02
improve docs

priyanshu450:


gabomdq:
Thanks for your contribution but this doesn't seem like a good fix.

PR #388, opened 2020-10-13
Add new 8BitDo Zero 2 gamepad for Linux

piot:


offalynne:
map's wrong, should be
`05000000c82d00003032000000010000,8BitDo Zero 2 gamepad,a:b1,b:b0,x:b4,y:b3,back:b10,start:b11,leftshoulder:b6,rightshoulder:b7,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,leftx:a0,lefty:a1,rightx:a2,righty:a3,platform:Linux,`

offalynne:
Which firmware version is this for? You should be able to map all of dpad/left stick/right stick as using Select + Up/Left/Right.

piot:
Yes, forgot about the special keys. Will update the PR soon.

gabomdq:
If you are going to update the PR, can you change the name to "8BitDo Zero 2" so it matches the other entries?

offalynne:
> Will update the PR soon.

@piot if/when you get around to this can you check to make sure the face button mappings are correct? I suspect this map is wrong because it differs from the other maps for the same device/platform: make sure to match ABXY to South/East/West/North like on Xbox 360 pads ignoring the labels on the actual controller.

PR #399, opened 2020-11-17
Added Ipega PG-9055 for Linux

pariziv:
Added Ipega PG-9055 for Linux


offalynne:
Redundant as GUID and map matches `Ipega PG-9069 - Bluetooth Gamepad`

gabomdq:
Dupe entry.

PR #414, opened 2020-12-29
Adding Ipega PG-9118

dm2912:
modified:   gamecontrollerdb.txt


offalynne:
are these working right with udev? would be curious to see how the buttons are being iterated with jstest.. `b87` ??

dm2912:
Tested in PPSSPP and Retroarch on Emuelec and all working fine and normally.  this controller is the bane of my life though.

hadogenes:
the left/right trigger are wrong, corrected in #461

PR #417, opened 2021-01-07
Add Saitek P220 Linux mapping

Ambrevar:
The Windows mapping was already added in https://github.com/gabomdq/SDL_GameControllerDB/pull/249.
Here I've used *trigger instead of *shoulder since it worked better for the one game I've tested it with (Streets of Rage 4).


Ambrevar:
Thanks!

PR #418, opened 2021-01-13
Add support for a PS3 gamepad clone

AmFobes:


offalynne (changes requested):
Map looks good, can you move the line up into the ‘platform:Linux’ section please?

PR #419, opened 2021-01-21
Added proper mapping for 8BitDo Wireless Adapter on Linux

mariotaku:
Please confirm before merge - DInput mode of this PR has A,B,X,Y reversed. On my Xbox One controller, they layout is
```
   Y
X     B
   A
```

I'm not sure previous version was a mistake, or it corresponds to different gamepads.


offalynne:
There have been multiple firmware updates at least one of which mentioned/addressed face button mapping.

I’m not sure if the dongle reorders the face buttons to match the labels or leaves them the same relative to cardinal direction (SDL and this repo defer to the latter).

I don’t have the hardware on hand to update/compare/check right now, but I can confirm in few weeks. If you want me to merge now and check later let me know and I’ll do so, otherwise just letting you know this has been seen! Thx for the PR

offalynne:
Checked with VM, looks good. Looks like the relevant firmware update was  12/11/2019 Firmware v1.30: 
> **2. Fixed the button mapping issue for A/B and X/Y for Xbox One & Xbox One Elite 2 on D-input mode.**

PR #421, opened 2021-01-22
add  MOCUTE-050X-M54-PC controller

Mrlinkwii:
add  MOCUTE-050X-M54-PC controller to Database   , mapping  gotten from https://github.com/PCSX2/pcsx2/issues/4163#issuecomment-763963487


offalynne:
AntiMicro uses an old version of the GUID format, please provide a map using a newer tool as indicated here https://github.com/gabomdq/SDL_GameControllerDB#mapping-tools

PR #422, opened 2021-01-22
Added OnyxSoft Dual JoyDivision and INNEX Genesis Controllers

midwan:
These are USB adapters that allow you to connect old-school 9-pin controllers, from Amiga, Atari, SEGA etc. 

The mappings were tested on Linux using Amiberry (Amiga emulator for ARM).


offalynne:
Do you think it's valuable to have mappings for these specific devices through these specific adapters? If different controllers were used with the same adapter device, would the layout possibly be different, and the name be incorrect? Are more buttons possible with other compatible pads? It seems the ambiguity of the device being used means providing a fixed map has more potential to be harmful than it would ever be helpful.

midwan:
The JoyDivision adapter is Amiga/CD32 specific, so any joystick/gamepad you connect will have the signals translated to the same standard from the controller's firmware. The mapping allows SDL2 to have a preset for that specific mapping, which is always the same, based on the original hardware specifications.

Amiga joysticks usually had 1 or 2 buttons (which was the maximum supported). CD32 pads came out a little later, and added more buttons (a DPAD, N/S/W/E, Start and 2 shoulder buttons). The controller's firmware expects such a controller, so the output is always going to be in a fixed mapping. Also, the name will always be the same, since it comes from the controller's firmware, not the connected joystick/gamepad.

I think it's helpful to have these, as it simplifies the mapping handling for all SDL2 apps using the GameController structure.

offalynne:
OK great, thanks for clarifying!

I am confused about the "INNEX Genesis Controller" is the name incorrect/nondescriptive of the actual device?

midwan:
That's just the name the firmware of the adapter reports to the system.
The actual connected joysticks/gamepads are never visible to SDL2 themselves, only the the adapter which handles the communication between them and the USB port (remember, these are old-school 9-pin "dumb" devices, they didn't contain modern logic).

The connections go like this: 
9-pin joystick -> USB controller/adapter -> USB port -> SDL2

So SDL2 sees only the events triggered from the USB controller/adapter, and can retrieve information like the device GUID and name from there. The controller handles that with a firmware which translates the 9-pin input accordingly, and presents itself as a "controller" to SDL2, sending the events as they come in.

Here's a link to the JoyDivision controller, to give you an idea: https://www.onyxsoft.se/joydivision.html
Hope that helps making more sense out of this...

offalynne:
If it's not actually an INNEX Genesis Controller, can you modify the name to better indicate the device?
> Please make sure to check that the name is a good description of the controller. If relevant, include the controller's name and model number.

midwan:
That is the name of that controller, it's not incorrect. It's what it reports back to the SDL2 controllermap tool.
The JoyDivision is another one, which again reports that name to the tool.

I could change it of course, but I just left what they reported themselves as names.

offalynne:
https://www.google.com/search?tbm=isch&q=INNEX%20Genesis%20Controller

Can you find a product/image link? The devices that match that name seem to have 6-8 buttons and you only have 2 mapped here.

midwan:
OK, let's remove the Innex Genesis one, since that is not Amiga specific (and not that important anyway). :)

offalynne:
Thanks for humoring me! I'll merge this now, but just for posterity sake, which firmware are you using on the JoyDivision adapter?
https://www.onyxsoft.se/joydivision.html I'm assuming you can only flash one at a time and you're using the Amiga CD32 fw?

midwan:
That's correct, the CD32 one is what makes more sense to have as a mapping anyway, otherwise you're just using 1-2 buttons which are always going to be Button0 and Button1.

PR #425, opened 2021-02-03
Add support for Sega Genesis Mini pads (Windows)

heyjoeway:
This adds Windows support for the official pads that ship with the Sega Genesis Mini.

Oddly, the d-pad is exposed as an analog input, though I've mapped it as a d-pad in this configuration. Other buttons are mapped as follows:
A -> Xbox A
B -> Xbox B
C -> Xbox X
Start -> Xbox Start

Generated using the SDL2 Gamepad Tool.


offalynne:
I know it’s a bit weird but I’m inclined to suggest C be mapped to RT as per the Hori Fightpad layout: https://images.app.goo.gl/hLGTzPjvozJP2AHr8

Since there’s a 6 button variant (which I’ll add a mapping for in follow-up) and that is generally the format we use for 6 button/Sega style pads.

This OK with you @heyjoeway ?

heyjoeway:
Yeah, looking at other examples, you're right. Just pushed the change!

PR #426, opened 2021-02-07
Fix invalid CSV due to using comma in controller name

akien-mga:
Fixup after ccd03220e80a0a367bd8d306015e28120a2b1702.


offalynne:
Thanks

PR #432, opened 2021-02-13
Update gamecontrollerdb.txt

suviljan:
I added Logitech Attack 3 on Windows and Linux.


offalynne:
This is a flightstick, correct?

suviljan:
Yes.

 - Susanna -

alynne o. kirjoitti 14/02/2021 klo 01:36:

offalynne:
This db is explicitly for gamepads and is meant to omit flightsticks. If it's mistakenly iterating as a gamepad with SDL you should take the issue upstream to SDL: https://github.com/libsdl-org/SDL

PR #441, opened 2021-03-07
change "NEXT SNES" right shoulder button from b6 to b5

tuket:
Hi,
I own the "NEXT SNES" controller and, aparently, all buttons were okay except for the right shoulder. I have tested this is Windows and Linux only, but I suppose it should be the same for other platforms.
Thanks!


offalynne:
@Superwaitsum @thecocce are these mappings valid for you?

offalynne:
Looks like it's `b5` and `b6` on different models. Could you add `righttrigger:b6` to each?

offalynne:
the format only allows one mapping per field, can you map to `righttrigger` instead svp ?

tuket:
Oh, I didn't know that you can't map to the same button twice. It seems to work though. But I only have one kind of controller to test (b5). Maybe it's the fist one that works? I'm gonna change the order and see what happens.

tuket:
Done!

> rightshoulder -> b5, righttrigger -> b6

I have also opened a discussion in the SDL forum if you are interested: https://discourse.libsdl.org/t/an-idea-to-change-the-behavior-of-sdl-gamecontrollergetbutton/30764

offalynne:
thanks !

PR #447, opened 2021-03-16
Update gamecontrollerdb.txt

RTHPJM:
Added 'GreenAsia Inc. USB Joystick' to Mac OS X section, UID 030000008f0e00000300000007010000


RTHPJM:
This is a USB dual stick controller, often sold in the UK by stores such as Aldi under their Tevion brandname

PR #455, opened 2021-04-10
Added ShanWan WUP-028 Gamecube Adapter

0verk1ll:
I set the configuration using AntiMicroX 3.1.5 and the physical switch on the adapter set to "PC" instead of "SWITCH/WIIU".

Micro Center carries this particular model of controller (https://www.microcenter.com/product/606375/ttx-tech-gamecube-controller-adapter-for-wii-u-and-switch).


offalynne:
How have you mapped the GameCube face buttons? Can you confirm (or edit so that) GameCube  B and X are SDL x and b? This is how these are meant to be laid out (matching location, not label). Thanks

offalynne:
@0verk1ll

0verk1ll:
Yes, I did map the buttons according to the labels on the face of the controller. Thank you for changing my commit.

offalynne:
thanks for confirming!

PR #456, opened 2021-04-10
I added the AntiMicroX Tool to the README

0verk1ll:
AntiMicroX is a GNU/Linux tool for generating SDL2 configurations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,9 @@ Please make sure to check that the name is a good description of the controller.
 ## Mapping Tools
 There are a few different tools that let you create mappings.
 
+### [AntiMicroX](https://github.com/AntiMicroX/antimicrox/)
+GUI tool for configuring controller profiles in GNU/Linux and generating SDL2 controller porfiles.
+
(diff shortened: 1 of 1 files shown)


offalynne:
We’ve got more than enough of these now and I’d much rather stick to platform-agnostic tools. The list in the Readme intends to be a recommendation, not an large list of alternatives.

Ryochan7:
> The database is compatible with SDL v2.0.10 and newer. Older SDL versions are no longer supported.

It does not look like that portion of the program has changed since I last worked on it. IIRC, the SDL GameController mapping strings generated by that window conformed to the standard around SDL 2.0.4. It works out well enough for the capabilities of the program but it cannot take advantage of some of the more advanced mapping options available in later versions of SDL.

offalynne:
@Ryochan7 yeah, I’m inclined to remove stuff that’s fallen out of date, I’d like to feature your Gitlab-hosted version of gamepad tool as our top recommendation and list the others as lesser alternatives — would you consider including Win/Mac/Linux binaries with your releases?

PR #465, opened 2021-05-14
WIP: Add Thrustmaster T150RS Racing Wheel - Linux platform

thomasa88:


thomasa88:
Hmm, should wheels be added or not? Will we lose force feedback?

offalynne:
This DB is explicitly for gamepads and deliberately excludes wheels and flightsticks. Use the more generalized sdl_joystick features.

PR #468, opened 2021-06-08
Add Raphnet Dual NES adapter for Linux

ilexiris:


offalynne:
@ilexiris can you confirm which is "B" and "A" on the NES controller using jstest or similar? they should be swapped in the mapping relative to Xbox controller buttons.

ilexiris:
@offalynne The NES-A is mapped to the Xbox-B, and the NES-B is mapped to the Xbox-A, in line with how SNES controllers are mapped.

PR #471, opened 2021-06-13
Update gamecontrollerdb.txt

Kohen-lab:
I added the mapping for the Magic-S pro from Mayflash as i noticed it was missing


offalynne:
Thanks!

PR #474, opened 2021-06-20
Remove entry for the teensy USB development board

natano:
GUID 03000000c01600008704000011010000 matches the teensy USB development board. I don't think having a mapping for it makes sense because the USB functionality can be freely programmed in the firmware and the listed configuration is only one of many possible configurations.

Having the (potentially) incorrect mapping in there can lead to problems in downstream projects, for example https://github.com/GrooveStats/gslauncher/issues/5.


offalynne:
Agreed, thanks!

PR #475, opened 2021-06-26
Added support for Nexilus GC adapter for Windows and Linux

WinterAlexander:
Support for this adapter on Windows and Linux 
https://www.amazon.ca/gp/product/B00S7CZWUQ/ref=ppx_yo_dt_b_asin_title_o01_s00?ie=UTF8&psc=1


offalynne:
How have you mapped A/B/X/Y ? Can you confirm they match based on relative position of the Xbox controller (so buttons labeled "B and X" are swapped) ?

PR #477, opened 2021-06-29
Updating PS4 Controller Mappings on Windows for new SDL2 versions

mpeterson-retrostudios:
The previous mappings resulted in incorrect behavior on SDL2 versions 2.0.14 and 2.0.15 (which Moonlight uses).

(This is also my first pr ever -- thanks for any and all help!)


offalynne:
@mpeterson-retrostudios can you confirm your device works correctly with SDL if this mapping is absent from the DB altogether ? see #478

mpeterson-retrostudios:
> @mpeterson-retrostudios can you confirm your device works correctly with SDL if this mapping is absent from the DB altogether ? see #478

Indeed it does -- it works correctly with SDL if this mapping is absent from the DB altogether.

offalynne:
Thanks for confirming,
closing as per #478

Issue #470, opened 2021-06-12
Per-platform maps for 8BitDo Lite, Arcade Stick, Pro 2, SN30 Pro for Android, and Modkits


offalynne:
@8bitdo

offalynne:
#473 adds a wide variety of these (thanks @8BitDo), would still like to see Arcade Stick, Pro 2 and Lite maps for Mac & Linux per mode and connectivity method

PR #482, opened 2021-07-21
Add 8BitDo Lite for Windows, Linux

shavs:
This merge adds the 8BitDo Lite controller configuration for Windows and Linux devices.

This follows from https://github.com/libretro/ludo/issues/342#issuecomment-883877082, and hopefully have a knock-on effect on [glfw's controller map](https://github.com/go-gl/glfw/blob/62640a716d485dcbf341a7c187227a4a99fb1eba/v3.3/glfw/glfw/src/mappings.h)

Please let me know if there is anything further that I can provide


shavs:
I should probably also mention that  #470 relates

offalynne:
Thank you so much for the PR! Unfortunately both of these GUIDs are in invalid -- the Windows map is being handled by SDL's own HID driver (suffix 7801) and the Linux map is identical to the Xbox One Wireless Controller. 

Most 8BitDo devices have unique fields even in Xbox spoofing mode that permit disambiguation on non-Windows platforms, but it seems like the Lite doesn't, at least on Linux. If you're able to get a unique GUID (not already in the DB) when wired, or in Switch mode, or are able to test further on Mac or Android please feel free to open another PR or follow up however suits you best!

Thanks again

shavs:
OK, thanks!

I did have a look at the difference in USB under Linux - it seems as though the controller is identified as an XBox 360 controller when plugged in via USB-C _and_ in Xinput mode.

I think, under "S" (Switch Pro mode), it'll emulate a Switch Pro controller, with seemingly the same GUID.

The only thing I haven't done normally during this, is use Steam Big Picture - does this have a bearing on the GUIDs generated?

This was the output of the `SDL_GamepadBind` section in `config.vdf`:

```
"SDL_GamepadBind"		"030000005e0400008e02000014010000,dev:gen:8BitDo Lite,platform:Linux,a:b1,b:b0,y:b2,x:b3,start:b7,guide:,back:b6,leftstick:b9,rightstick:b10,leftshoulder:a2,rightshoulder:a5,dpup:h0.1,dpleft:h0.8,dpdown:h0.4,dpright:h0.2,leftx:a1,lefty:a0,rightx:a3,righty:a4,lefttrigger:b4,righttrigger:b5,
...
```

I edited the result to match that of the database (removing `dev:gen` (their templating for the device - `device: Generic Gamepad`, it can also be `dev:xb1` for Xbox One-based gamepads).

Would the result differ in a different tool, or would it still result in the same GUIDs?

Thanks again! :smile:

offalynne:
Thanks for posting the Big Picture map, that too is a duplicate GUID. 

> Would the result differ in a different tool, or would it still result in the same GUIDs?
Depending on which version of SDL the tool is built with the result may vary, but in most circumstances the GUID should be the same.

PR #487, opened 2021-08-04
Added "PowerA Wired Controller for Xbox One" (Windows)

Zbyl:


cgutman:
This is a mapping for an XInput controller. The built in SDL should handle it already. Did you find that not to be the case?

offalynne:
🚨`[…]00007801`🚨

thanks @cgutman. any chance moonlight project folk wanna fork ryochan7’s updated tool on gitlab and distribute up to date binaries for our benefit? 😉

PR #493, opened 2021-08-08
Add another PS4 controller variant for Android

JujuAdams:


offalynne:
Thanks! I’ve got this & a dozen other odd maps I’ve been sitting hesitant on; this is the Old Jank GUID Format that SDL doesn’t use anymore (ofc our good friends YYG do).

Issue #478, opened 2021-06-29
Built-in SDL HIDAPI/XInput mappings are being overridden


cgutman:
This is a continuation of the discussion from 9219d19540747efca0019da69064fc65cf000db0, https://github.com/moonlight-stream/moonlight-qt/issues/620, and #477. I took a look at the SDL code to figure out what's going on.

Some of the mapping GUIDs have special suffixes that mean they are associated with a particular API:
https://github.com/libsdl-org/SDL/blob/c59d4dcd38c382a1e9b69b053756f1139a861574/src/joystick/windows/SDL_xinputjoystick.c#L279-L280
https://github.com/libsdl-org/SDL/blob/b63cb822bf02ffa53e7a25755ff4606af0417389/src/joystick/hidapi/SDL_hidapijoystick.c#L918-L920

The trouble is that these special APIs almost always have an existing mapping built into SDL like [this one for HIDAPI](https://github.com/libsdl-org/SDL/blob/afd100f02b7689e3b1cf441f7abf40f0313f328b/src/joystick/SDL_gamecontroller.c#L574). Allowing mappings for these special APIs here creates the possibility for error like we've seen here.

For HIDAPI joysticks, this suffix is `6800` (0x68 = 'h') which is exactly what the suffix of the broken mapping is from 9219d19540747efca0019da69064fc65cf000db0. The mistake there seems to be that they assumed it would be equivalent to the non-HIDAPI mapping for a PS4 Controller. This is not the case, since HIDAPI does not necessarily expose all axes and buttons in the same order that DirectInput does.

[...]

PR #498, opened 2021-08-20
Update gamecontrollerdb.txt

kode54:
Added Xbox Series X controller for macOS, using
SDL2's official mapping tool to map all supported
controls


offalynne:
Hi @kode54 

Can you check if the device works OK in SDL *without* this map? The `6800` at the tail of the GUID suggests it's already being handled internally by SDL, see #478

kode54:
You are correct, it appears to be natively supported by the latest SDL2.

PR #505, opened 2021-10-07
Add Nintendo Switch PowerA Wired Controller on Linux

CallumDev:
Mappings auto-detected are incorrect otherwise


offalynne:
Have you tested this thoroughly? The right stick horizontal axis mapping looks pretty weird.

Issue #507, opened 2021-10-12
Add Thrustmaster T.Flight HOTAS X controller


AgentSmith0:
Here are the mapping strings for the Thrustmaster T.Flight HOTAS X controller (https://support.thrustmaster.com/en/product/tflighthotasx-en/):

Linux: `030000004f04000008b1000000010000,Thrustmaster T.Flight Hotas X,platform:Linux,a:b7,b:b6,x:b5,y:b4,back:b10,start:b11,leftstick:+a0,rightstick:a2~,leftshoulder:b2,rightshoulder:b3,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,-leftx:-a0,lefty:a1,rightx:a4,lefttrigger:b1,righttrigger:b0,`

And also works on Windows: `030000004f04000008b1000000010000,Thrustmaster T.Flight Hotas X,platform:Windows,a:b7,b:b6,x:b5,y:b4,back:b10,start:b11,leftstick:+a0,rightstick:a2~,leftshoulder:b2,rightshoulder:b3,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,-leftx:-a0,lefty:a1,rightx:a4,lefttrigger:b1,righttrigger:b0,`

PR #508, opened 2021-10-12
Add Thrustmaster T.Flight HOTAS X controller for Linux and Windows

AgentSmith0:
Fixes https://github.com/gabomdq/SDL_GameControllerDB/issues/507


offalynne:
This DB is for *gamepads* and like devices. Flight sticks are deliberately omitted and are covered in SDL by the general unmapped case.

PR #515, opened 2021-11-06
Fix Nintendo GameCube Controller using official USB adapter on Windows

Superstarxalien:
Fixes the DInput mapping for Nintendo GameCube controllers on Windows using elmassivo's vJoy-based drivers.

Nintendo GameCube controllers do not have left shoulder buttons or back buttons. My fix also maps the B button on the GameCube controller to the X button on an Xbox controller.


offalynne:
To be honest I don't think the original map should really be in here in the first place except *possibly* for the default configuration, if vJoy ships one. People use the software-driven driver for all kinds of stuff and there's no one "correct" mapping.

As for the official USB GameCube adapter, SDL ships a driver for it.

offalynne:
opinion @cgutman ?

cgutman:
Yeah, the mapping in there is for the default vJoy configuration. We definitely shouldn't be changing that.

There needs to be a way to distinguish one type of vJoy device from another before we can add different mappings for them. If people are buildings drivers on top of vJoy, they should probably get some value reserved in the PID space for vJoy (which itself seems made up anyway: 0x1234). That would allow game engines and mapping databases like this one to be able to support the drivers out of the box.

PR #521, opened 2021-12-06
PBTails Gamepad (Nintendo Switch Pro Controller)

samermurad:
So there is this controller: https://www.pbtails.com/

Which Identifies itself as "Nintendo Switch Pro Controller", the specification of the layout looks like so:

Joystick Name: Nintendo Switch Pro Controller
GUID: 030000007e0500000920000001006800
6 axes (RT and LT are read as an axis)
0 trackballs
0 hat(s)
16 buttons

I've tested it only on macOS, so I don't know if it gets the same Id on all platforms?

The controller still works without the specification, but the X/O  Triangle/Square behavior is flipped (Nintendo style), this specification gives the expected result


offalynne:
Thanks for the PR.

The GUID tail suggests the DB shouldn't need this mapping since the device is handled by SDL's HIDAPI mapping: https://github.com/gabomdq/SDL_GameControllerDB/pull/521

> The controller still works without the specification, but the X/O Triangle/Square behavior is flipped (Nintendo style), this specification gives the expected result

I don't exactly understand the problem you're having here, but since the device reports the same as the Switch Pro controller, you'll have to take the mapping issue [upstream to SDL](https://github.com/libsdl-org/SDL/) if it needs addressing.

PR #523, opened 2021-12-07
add retro-bit 2.4GHz pads

caio2k:
Hello, I'm trying to add 2 pads with 2 modes each at the same time, please let me know if I've done it right.


offalynne:
Maps already in the DB for these, but thanks !

PR #526, opened 2021-12-14
Rename Tomee Controller

willco007:
Removes non-existent buttons from the Tomee SNES controller, fixes A mapping for macOS


offalynne:
The SNES controller doesn't have X Y or L R buttons ??

willco007:
This appears to be mis-labeled as well, it's the NES controller, not SNES. It's this:

https://www.amazon.com/Tomee-NES-Controller-nintendo-entertainment-system/dp/B0034ZOAKO/ref=sr_1_4?keywords=Tomee+NES+USB&qid=1639441703&sr=8-4

offalynne:
Looks like they share a GUID, and since this map is a subset of the one authored for the SNES model it's preferable to use the old map. @andrewhavens can you confirm if able ?

willco007:
I'm not surprised in the slightest they would share a GUID. I defer to your judgement on this one.

PR #527, opened 2021-12-14
adding ASCII Seamic controller and Hori Fighting Commander 3

zelurker:
ASCII Seamic : 1 stick
Hori Fighting Commander 3 : 3 sticks
windows only

It comes from a user, raine emulator.

### DIFF
--- a/gamecontrollerdb.txt
+++ b/gamecontrollerdb.txt
@@ -2,6 +2,8 @@
 # Source: https://github.com/gabomdq/SDL_GameControllerDB
 
 # Windows
+030000000d0f00002500000000000000,10B,platform:Windows,a:b1,b:b2,x:b0,y:b3,back:b8,guide:b12,start:b9,leftshoulder:b4,rightshoulder:b5,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,lefttrigger:b6,righttrigger:b7,
+03000000490b00004406000000000000,ASCII Mic/Joy-stick,platform:Windows,a:b0,b:b1,x:b3,y:b4,back:b8,start:b9,leftshoulder:b5,rightshoulder:b2,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,leftx:a0,lefty:a1,lefttrigger:b6,righttrigger:b7,
 03000000fa2d00000100000000000000,3DRUDDER,leftx:a0,lefty:a1,rightx:a5,righty:a2,platform:Windows,
 03000000d0160000040d000000000000,4Play,a:b1,b:b3,back:b4,dpdown:b11,dpleft:b12,dpright:b13,dpup:b10,leftshoulder:b6,leftstick:b14,lefttrigger:b8,leftx:a0,lefty:a1,rightshoulder:b7,rightstick:b15,righttrigger:b9,rightx:a3,righty:a4,start:b5,x:b0,y:b2,platform:Windows,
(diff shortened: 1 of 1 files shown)


offalynne:
Hi, thank you for these! Can you please refer to the mapping guide for button layout to ensure the mappings are correct? 
https://github.com/gabomdq/SDL_GameControllerDB#mapping-guide

Thanks again.

zelurker:
Ok, finally updated the mapping for the ascii pad, sorry for the delay, a bit of misunderstanding here.
By the way, the hori controller actually has no sticks which makes a it a little extreme for these mappings, but it's still useful to recognize the d-pad.
Anyway that's the best we can do here !

zelurker:
Oh by the way, if you are interested, here is the photo of the ASCII Seamic controller : https://segaretro.org/Seamic_Controller
And here is the hori fighting commander 3 : https://i.ebayimg.com/d/w1600/pict/283353630372_/SONY-PS3-Controller-Hori-Fighting-Commander-3-Black.jpg

offalynne:
@zelurker not too sure on these.. could you please ask your user to refer to the [Windows game controller dialogue](https://www.top-password.com/blog/open-game-controllers-settings-in-windows-10/) and make careful note of what each of the buttons are registering to be sure ? for example A = 3, B = 5, C = 7, etc. thanks !

zelurker:
Errr, it's really becoming complicated just for 2 small mappings... Yeah
one of the mapping follows the sega controller on your picture because it
looks like a sega controller (same disposition for the buttons), and the
other follows the playstation convention (obviously on the picture since it
uses the same playstation logos on the buttons). I'll forward the message a
last time, but if you prefer to refuse the PR, just let me know, it will be
faster and easier !

offalynne:
> it's really becoming complicated

Sorry for the trouble but thank you for considering my request. We do our best to try and provide good oversight here for each and every mapping to the best of our ability so players have a good experience with their devices.

> one of the mapping follows the sega controller on your picture because it
looks like a sega controller (same disposition for the buttons), and the
other follows the playstation convention

This tracks with my expectation, thank you. Looking carefully at the ASCII Seamic mapping I am not confident your user has correctly configured the buttons C, Z, L and R as per the Sega spec. If you could ask them to provide notes on the DInput (Windows Game Controller dialogue) results for those buttons, I would appreciate it. No hurry.

dezraj:
Hello! I am the owner of these pads. I mapped my ASCII pad based on the "Sega -> Microsoft" layout provided in this database. So when the SDL mapping program asked for "R trigger" and the "R (shoulder) button", they were respectively mapped to C and Z buttons on the ASCII pad. Perhaps you are in doubt because the Windows Game Controllers properties counts the buttons from 1, whereas the SDL program which I used counts them from 0. Anyway, here you have what the Windows reports:

ASCII Seamic Controller:

A = 1
B = 2
C = 3
X = 4
Y = 5
Z = 6
L = 7
R = 8
Select = 9
Start = 10

Hori Fighting Commander 3:

Square = 1
Cross = 2
Circle = 3
Triangle = 4
Left Shoulder = 5
Right Shoulder = 6
Left Trigger = 7
Right Trigger = 8
Select = 9
Start = 10
PlayStation Button = 13

What's odd about the Hori pad is that it doesn't start counting from the the X/Cross button, but from the Square one. Also, the driver reports 13 buttons but buttons 11 and 12 don't exist. As a side-effect the PS button becomes number 13.

Anyway, I hope this can help. 👍

offalynne:
@dezraj Thank you for the thorough follow up. 

> Windows Game Controllers properties counts the buttons from 1 SDL program counts them from 0
> Hori pad is [..] doesn't start counting from the the X/Cross button, but from the Square
> driver reports 13 buttons but buttons 11 and 12 don't exist
> PS button becomes number 13

These are all as expected, no issue. Thanks !

PR #528, opened 2021-12-19
Mapping for Xbox One S and Series controller on firmware 5.11.3118.0

XirXes:
These are needed to use the newest controller firmware with the Xbox Wireless controller adapter with the experimental branch of xone.


offalynne:
Thanks!

PR #529, opened 2021-12-20
add support for Tomee snes adaptor on linux

nstbayless:
note that Tomee snes adaptor reports its name as "SNES PC Game Pad    SNES PC Game Pad   "

Should the name be changed to "Tomee SNES Game Pad" or something like that? Or would that cause the entry to no longer match against the gamepad?


offalynne:
Devices match on GUID only, feel free to set the name to something appropriate. From the Readme:
> Please make sure to check that the name is a good description of the controller. If relevant, include the controller's name and model number.

nstbayless:
> @nstbayless please test this mapping and confirm it works correctly, I’ve swapped the axes to match other similar maps for the same brand of device.

@offalynne, what's the best way to test the db? Is there a simple test dialogue I can use?

offalynne:
@nstbayless How did you build the map initially? The ReadMe contains a few different tools you can try.

nstbayless:
@offalynne I have now tested and confirmed that your edit works using https://generalarcade.com/gamepadtool/

offalynne:
Thanks!

PR #535, opened 2022-01-08
FIX PS360 Windows

kab909:
Resolve conflict: value "h0.4" was set for 2 keys


offalynne:
Do you own the device this map matches ? Curious if you can provide a better device name.

PR #539, opened 2022-01-14
Added mapping for PowerA Core Plus controller on Linux

manadream:
I had two controllers that had this GUID, which were not in the DB, so I added the mapping.


offalynne:
Thank you !

PR #543, opened 2022-01-26
Added variant controllers seen using Antstream Arcade 

CdrJamison:
Added various controllers used by players of Antstream Arcade which didn't have entries in the database.


offalynne:
Thanks !

PR #544, opened 2022-01-26
Xbox Series X Controller,platform:Linux

paroj:

PR #545, opened 2022-01-27
add github workflow to detect duplicates

paroj:
sample output for #544
https://github.com/paroj/SDL_GameControllerDB/runs/4966298974?check_suite_focus=true

if you set CROSS_PLATFORM, it will help catching inconsistent naming, e.g.
```
Duplicate: 8Bitdo SFC30 GamePad Joystick at line 726
         -8Bitdo SFC30 GamePad +8Bitdo SFC30 GamePad Joystick -platform:Windows +platform:Mac OS X
```

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,0 +1,11 @@
+name: Data check
+on:
+  push:
+    branches: [master]
+jobs:
+  duplicates:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+    - run: |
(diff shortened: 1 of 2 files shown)


offalynne:
Neat. We did have a map validation script in the past (I think the SDL project might still ship an old one?) but it was lost in an attribution dispute, https://github.com/gabomdq/SDL_GameControllerDB/blob/master/check.py.

Leaving this up to the discretion of @gabomdq.

gabomdq:
Thanks!

PR #547, opened 2022-01-31
Add 8BitDo Pro 2 mappings for Linux

zephryn:
Only DirectInput mode appears to produce unique GUIDs. Connections under Bluetooth and USB produce different GUIDs, `05000000c82d00000660000000010000` and `03000000c82d00000660000011010000` respectively.

My gamepad's firmware version is 1.05.

These mappings use axes for the L2/R2 analog triggers (a5, a4) instead of the digital button inputs (b8, b9). Mappings were generated using SDL2 Gamepad Mapper AppImage.


offalynne:
Thank you !

PR #549, opened 2022-02-07
Linux: hid-nintendo Joy-Con fixes

HTV04:
Button mapping fixes for standalone Joy-Cons using the hid-nintendo module for Linux.

For some reason the right Joy-Con was almost mapped correctly, requiring only a few minor changes. However, the left joy-con was mapped completely wrong.


offalynne:
Is this with the latest version of the driver? These maps were mine but the driver has seen some revision and I wasn't aware it had made it into the kernel yet.

offalynne:
Why is there a `guide` field on the left JoyCon?

HTV04:
> Why is there a `guide` field on the left JoyCon?

I mapped the screenshot and home buttons to the guide button because I thought they made more sense. I can revert the change if you want, though.

offalynne:
Actually for parity with other solo joycon mappings those (Capture, Home) should be `back`. If you make that edits and can clarify which version of the driver you’re on I’ll be glad to merge.

HTV04:
> For parity with other solo Joy-Con mappings please set “-“ and “+” to `start` and “Capture” and “Home” to `back`. If you make the necessary changes and can clarify which version of the driver you’re on I’ll be glad to merge.
> 
> @HTV04

Just pushed a commit that applies these fixes. I'm not sure what version the hid-nintendo module is since it's built into my kernel, but here's my kernel version: 5.16.7-arch1-1

offalynne:
thanks !

PR #551, opened 2022-02-15
Add Xbox Wireless Controller (Daystrike Camo Special Edition)

lazd:
I got this sweet new red camo Xbox controller and was surprised to find it was not mapped correctly for SDL2 games, specifically testing on RetroPie.

066fcbfe-bfb3-488e-a5b9-fc8c1c5dc0ab

Xbox Wireless Controller, Daystrike Camo Special Edition, M1139678-001

I used SDL2's `controllermap` application to create a mapping and found it working perfectly thereafter.


offalynne:
Thanks !

PR #554, opened 2022-03-02
Add Atari VCS Modern Controller

ooshlablu:
This config allows the [Atari VCS Modern Controller](https://atarivcs.com/atari-vcs-wireless-modern-controller/) to be detected and automatically configured in SDL2.

FWIW, I used https://gitlab.com/ryochan7/sdl2-gamepad-mapper to get the mapping. I did so on an Ubuntu 20.04 system using the AppImage released about a year ago: https://gitlab.com/ryochan7/sdl2-gamepad-mapper/-/releases


offalynne:
Thank you!

PR #555, opened 2022-03-12
Add Xbox One mapping with wireless adaptor dongle

jgabriel98:
This provides mapping for Microsoft Xbox one controller when connected via dongle (microsoft wireless adaptor).

This mapping string was generated with [SDL2 Gamepad Mapper](https://gitlab.com/ryochan7/sdl2-gamepad-mapper)

## Context:
The dongle support was achieved with [xone](https://github.com/medusalix/xone) incredible work. But it seems the _gamepad GUID_ is different when using the dongle VS conected by cable.


jgabriel98:
I could not think of a better name for the device, since i seems to be the first "device connected by dongle" added.

So i'm accepting suggestions

offalynne:
Thanks, no need to specify 'adapter' here, I think.

PR #556, opened 2022-03-13
Add PDP Xbox 360 Wired

RedDevilus:
Adds Mac variant for PDP Xbox 360 Wired.
Also re-order entry to be alphabetically.


offalynne:
The values on this map are a total mess, please prompt your user that submitted this to retry setting the mapping and properly test them before submitting.

offalynne:
Duplicate

PR #560, opened 2022-03-29
Support Xbox Wireless Controller (model 1914) v5.13 on Linux

tchebb:
There is a new firmware update for the Xbox Wireless Controller for Xbox Series X|S (model 1914). The version mine is running is v5.13.3143.0. Add mappings for this firmware over both USB and Bluetooth.

These mappings were generated fresh using `controllermap` and reordered to match existing mappings as closely as possible. I changed the detected name of "Xbox Series X Controller" to "Xbox Series Controller" for consistency with others and to indicate that this controller comes with both the Series X and the Series S.


offalynne:
Thank you !

PR #564, opened 2022-04-14
Linux: Improve Mayflash GCN Mapping

flagrama:
This is probably going to need a bit of discussion. The GameCube controller is awkward. It obviously has the triggers that work as you'd expect, but if you've used one you know there is a click at the end of the axes. This click is a separate button. This is often overlooked.

This mapping remaps these buttons to be the shoulder buttons. This means the single right shoulder button (Z) needs a different mapping. I have mapped it to "back" as that seems like the least confusing of options given the aforementioned shoulder button situation, however I am up to suggestions.

I believe that this mapping should work as is if copied to the Windows GUID for this controller as well, the button mappings seem equivalent, and the existing mapping seems to map multiple buttons to a single button? However, I'm unable to test this so have not included it.


offalynne:
Thanks for the PR! Coincidentally I picked up the adapter myself and intend to clean up the mappings cross platform.

Here’s the GameCube mapping for face buttons
https://github.com/gabomdq/SDL_GameControllerDB/blob/master/mapping_guide.png

L and R should be mapped to analogue pull if possible. The full-pull trigger button does not need to be mapped for the purpose of SDL support (Xbox-like compatibility).


Z should be right shoulder shoulder. The back, guide, and left/right stick mappings should be left blank since the controller lacks them.

I understand the merit of repurposing the RetroArch retropad mappings in order to fit necessary extra features into the map, but this is not a good idea for SDL, which aims primarily to conform mappings to XInput style.

flagrama:
With a GameCube controller you're really only going to have triggers or shoulder buttons, never both, anyway and personally I like having the option between using a button or using the triggers.

At least on my device or controller, my triggers are immensely sensitive. Using them as triggers is generally fine, but trying to use them as buttons just results in them constantly being "pressed". Even making this mapping using the SDL Gamepad Mapper from the readme, I had to manually enter A, B, Left Trigger, and Right Trigger after the fact because A and B were always a3 and a4 since otherwise I'd have to fight the triggers through half of the mapping process. I don't really find it a good user experience.

On the other hand, I've got this mapping now and I can just manually use it so that I can avoid that personally. It could also just be an issue with my hardware, in which case it wouldn't affect the normal user and using this mapping personally would be the correct choice anyway.

PR #571, opened 2022-05-09
Add Nacon GC-100 (Windows)

Jeronimo17:


offalynne:
@Jeronimo17 thanks for the PR! Because the GUID ends in `7801` we can assume SDL is already handling this device correctly per #478

Jeronimo17:
Thanks to you, in windows if I have seen that it works well I have only had problems with 1 game, in linux adding it is when it has started to work with almost all games.

PR #578, opened 2022-05-30
Added MSI GC20 V2 (variant).

fopenp:
I previously sent an MSI GC20 configuration taken from Steam. From another dev software (Godot Joypad Demo Tool) I had another serial and a slightly different configuration. This one I manually remapped the buttons, so I'm sure it's ok. I suggest you to keep both.


fopenp:
I seen a button used twice. Maybe the program has some issue.

PR #582, opened 2022-06-10
Added another Hori Real Arcade Pro 3 controller

nightmareci:
The Hori Real Arcade Pro 3 I have has a Linux GUID different than the one that was already in the database.


offalynne:
Thank you

PR #583, opened 2022-06-10
adds Tomee NES & SNES controller adapters

stevenjamescox:
Adds Tomee NES & SNES controller adapters for macOS. They look like this:

hyperkin_m07020_tomee_nes_controller_to_1359838 hyperkin-tomee-nintendo-snes-to-usb-controller-adapter-340015


offalynne:
Thank you very much !

Issue #202, opened 2018-03-28
Duplicated/conflicting GUIDs for the DragonRise/PC TWIN SHOCK controllers


Swyter:
Turns out that many generic Chinese controllers based on the same _PC TWIN SHOCK_ board from _DragonRise Inc._ share the same USB vendor (`0x0079`) and product (`0x0006`) identifiers, but they also use slightly different button layouts, causing conflicts because the SDL2 gamepad GUID cannot distinguish between them.

My _NGS Phantom_ mainly conflicts with #86 in Linux, and the "_G-Shark GS-GP702_" in Windows. Funnily enough in macOS looks like the GUID is slightly different, for some reason.

These are the correct mappings for Linux, macOS and Windows:

```
03000000790000000600000010010000,NGS Phantom---LNX,a:b2,b:b3,x:b0,y:b1,back:b8,start:b9,leftstick:b10,rightstick:b11,leftshoulder:b4,rightshoulder:b5,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,leftx:a0,lefty:a1,rightx:a3,righty:a4,lefttrigger:b6,righttrigger:b7,platform:Linux,
03000000790000000600000007010000,NGS Phantom---MAC,a:b2,b:b3,x:b0,y:b1,back:b8,start:b9,leftstick:b10,rightstick:b11,leftshoulder:b4,rightshoulder:b5,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,leftx:a0,lefty:a1,rightx:a3,righty:a4,lefttrigger:b6,righttrigger:b7,platform:Mac OS X,
03000000790000000600000000000000,NGS Phantom---WIN,a:b2,b:b3,x:b0,y:b1,back:b8,start:b9,leftstick:b10,rightstick:b11,leftshoulder:b4,rightshoulder:b5,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,leftx:a0,lefty:a1,rightx:a2,righty:a4,lefttrigger:b6,righttrigger:b7,platform:Windows,
```

I ship a game with these community mappings and I'm having trouble playing with my own controller because everything is subtly broken.

_PS_: Maybe Ryan C. Gordon can add some additional optional bits to tell them apart. Maybe by optionally matching by VID/PID/Version/Revision/Firmware instead of only VID/PID.

[...]

PR #584, opened 2022-06-11
adds 8 controller entries for macOS

stevenjamescox:
adds the following controller entries for macOS:
- Hyperkin N64 Controller Adapter
  - used to connect a standard N64 controller to a device via USB 
  - one entry is for PC Mode, the other for Console Mode (there's a small toggle switch on the adapter)
- Unbranded/Generic N64 USB Controller
  - apparently its hardware device IDs (vendor 121, product 6) are used by a variety of generic USB controllers out there, but its addition doesn't create any SDL GUID conflicts
- Sega Genesis Mini Official Controller
- 8BitDo M30 2.4 GHz for Sega Genesis Mini
- SteelSeries Stratus Duo
  - one entry is for utilizing its included 2.4 GHz USB receiver, the other is for connecting via a micro USB cable
  - in both such modes it operates as an Xinput controller and thus requires a modified 360Controller.kext (or other such driver) to actually use, but such is also the case with several other controllers currently in the database under macOS so it did not seem inappropriate to add them
  - this controller is also capable of connecting via Bluetooth, but will not stay legitimately connected in macOS
- Retro-Bit Sega Genesis 6-button Arcade Pad with USB


offalynne:
Thank you for taking the time to find missing devices and mapping them carefully and correctly!

Unfortunately the ubranded N64 pad does not look to be reliably identifiable on that GUID alone as even the version/bcdDevice field is not unique: https://github.com/gabomdq/SDL_GameControllerDB/issues/202#:~:text=%3E%20%20%20bcdDevice%20%20%20%20%20%20%20%20%20%20%20%201.09

I've removed that mapping but otherwise looks good. Will merge shortly.

stevenjamescox:
My pleasure. Thank you for cleaning 'em up!

PR #585, opened 2022-06-12
Fix iBuffalo SNES Controller mappings (Windows)

GhostlyDark:
Confused about the previous values. This controller doesn't have triggers and the shoulder button values didn't match. X and Y as well as A and B were swapped, now matching Xbox button layout.


offalynne:
A number of vendors including Retro Bit and Retro Fighters make Saturn-style devices that spoof the iBuffalo SNES style pad. If you compare closely you’ll notice the mapping you overwrote is a superset of the one you submitted, laid out in the “Sega” format from the README. The face button pairs are swapped on those devices as well — if you use gamepad type identification in your software designating the device ID as “like Saturn” and use an appropriate button label to input lookup, the other mapping will work correctly for all devices instead of just one.

It’s not ideal, but I think it’s an acceptable compromise. PRs from other DB users in the past have suggested we always defer to whichever device has more buttons which I don’t necessarily agree with that, but unless one layout represents a device with a vastly larger market share whose mapping is entirely different, I think this kind of compromise is best.

PR #586, opened 2022-06-14
standardization/normalization/fixes for some SteelSeries controllers

stevenjamescox:
- SteelSeries Nimbus for macOS `b12` mapping consistency:
  - This controller unfortunately does not have the "normal" set of three buttons in the center. Just one:
steelseries-nimbus-wireless-controller-335759
  1. `03000000110100002014000000000000` used `b12` for `start`
  2. `03000000110100002014000001000000` used `b12` for `guide`
  3. `03000000381000002014000001000000` used `b12` for `guide`
^I changed the latter two entries' `b12` elements to `start`.
Despite that this button functions more like `guide` in iOS and tvOS, setting it to `start` in the database seemed to better match how other controllers with lone central buttons are configured (N64, Sega Genesis, et al.). If this is incorrect, then it would follow that the first of the three entries needs to have its `b12` changed to `guide` instead.
  
- SteelSeries Nimbus for macOS y-axis mapping consistency:
  1. `03000000110100002014000000000000` used `a1`   for `lefty` and `a3`   for `righty`
  2. `03000000110100002014000001000000` used `a1~` for `lefty` and `a3~` for `righty`
  3. `03000000381000002014000001000000` used `a1~` for `lefty` and `a3~` for `righty`
^I changed the first entry's y-axis elements to `a1~` and `a3~`, respectively.
I have this controller. In macOS, it reports the second above-listed GUID when connected via Bluetooth and the third when connected via a lightning-to-USB cable. Maybe the first GUID above is a rare variant. Or maybe it's a Nimbus that has been run through an intermediate driver/application of some sort. I'm not sure. If it is a correct direct mapping, it would be truly strange as all other mFI controllers entries in the database (Nimbus, Nimbus+, Stratus XL) have both of their vertical axes "inverted" by default in macOS and thus have `~` in their mappings. It's _most likely_ a misconfiguration. I may be wrong, though.

- SteelSeries Stratus Duo for Windows:
  - `03000000381000003014000000000000` - prepended 'SteelSeries' to controller name
  - `03000000381000003114000000000000` - prepended 'SteelSeries' to controller name

- SteelSeries Nimbus+ for Linux:
  - `050000004e696d6275732b0000000000` - prepended 'SteelSeries' to controller name


offalynne:
Can you revert the `guide -> start` change ? While I agree with you in theory (the start mapping makes the device more immediately useful in games) the goal of the DB is normalizing to device label, and so the `guide` mapping is correct. 

If you prefer the `start` mapping for your own usage, you can use the mapping tool to set an environment variable which should work for local SDL/Steam games.

offalynne:
Thank you for the PR !

PR #588, opened 2022-06-16
some SteelSeries additions & fixes for Windows

stevenjamescox:
- Added `030000003810000060b3000000007801` - SteelSeries Nimbus for Windows - This controller requires [MFI Gamepad Feeder](https://axadiw.github.io/MFIGamepadFeeder/) (or a similar driver if it exists) for it to have _any_ function in Windows.

- Added `030000000d0000000000000000000000` - SteelSeries Nimbus+ for Windows - R2 & L2 are missing because they're not recognized by SDL2 Gamepad Tool. They aren't recognized in RetroArch or any other application I tried. But they _do_ appear to work in Windows Control Panel's "HID-compliant game controller" Properties interface.

- No change to `03000000110100003114000000000000` - SteelSeries Stratus Duo via Bluetooth for Windows - The guide button is missing but it's not recognized by SDL Gamepad Tool. It makes sense that this button doesn't function well in Windows as this controller's Bluetooth mode was specifically designed for Android.

- Added missing guide button to `03000000381000003014000000000000`- SteelSeries Stratus Duo via 2.4GHz USB receiver for Windows - The guide button will bring up the Xbox Game Bar by default, but this can be easily turned off in Settings by unchecking a box in the section for Xbox Game Bar. It appears that this box must be unchecked for the SDL2 Gamepad Tool to be able to record the guide button input.

- Added missing guide button for `03000000381000003114000000000000`- SteelSeries Stratus Duo via USB cable for Windows - Same situation as the previous one.

- Added `03000000381000003014000000007801` - SteelSeries Stratus Duo via 2.4GHz USB receiver for Windows - Maybe this a legitimate alternate GUID, but I'm thinking it's just a duplicate of `03000000381000003014000000000000` because I'm not seeing `7801` at the end of any current Windows GUIDs. This is, however, the number my controller brings up in SDL2 Gamepad Tool. Could this have anything to do with the tool being slightly outdated and/or using a less-than-current version of SDL?

- Added `03000000381000003114000000007801` -  SteelSeries Stratus Duo via USB cable for Windows - Same situation as previous one, this is probably an erroneous duplicate.


offalynne:
The `7801` suffix indicates XInput, which means a device should not need mapping as SDL handles it already: https://github.com/gabomdq/SDL_GameControllerDB/issues/478#issuecomment-870985088

`030000000d0000000000000000000000`: This GUID is rather the opposite of Unique and doesn't inspire confidence, so I think it's best to omit this as well. Less of a problem for other mapping projects like RetroArch that can match on name alone.

All good as for the rest, thank you for the helpful detail!

stevenjamescox:
Thank you, as always, for your patience with my being a novice!

I keep seeing devices' hardware IDs split in half and reassembled back-to-back near the middle of these GUIDs, is that standard? I've been trying to find some sort of reference material that fully explains how to break down SDL GUIDs but have come up empty. 

The Nimbus+ is truly an odd one. Vendor ID 0x000d, Product ID **0x0000**. I am not familiar with the particulars of HID-compliance, device-identification compliance, etc., but it seems like a product ID of zero has to be a no-no. It's like they thought, "This'll just be used on iOS/tvOS, so we _only_ have to care about its adherence to MFi specs," OR could it be that it _doesn't even have_ a product ID and so systems just show it as zero?

offalynne:
GUID format is
```
XXXX 0000 XXXX 0000 XXXX 0000 XXXX XXXX
|    |    |    |    |    |    |    |
|    |    |    |    |    |    |    Flag
|    |    |    |    |    |    Version
|    |    |    |    |    Padding
|    |    |    |    Product
|    |    |    Padding
|    |    Vendor
|    Padding
DeviceBusType
``

Where Flag is one of several internal values: https://github.com/libsdl-org/SDL/blob/d58d637ac6bd84a3fb30ecbfc50e8454fac9d9fa/src/joystick/SDL_joystick.c#L2190-L2217

> split in half and reassembled back-to-back

This is just an endianness difference

> reference material

Not much available, pretty much have to read the C source

Very few of these gamepad devices (including all of first-party console manufacture) are properly HID-complaint, it does seem like most vendors use the spec to the bare minimum necessary to get things working and no further.

> could it be that it doesn't even have a product ID and so systems just show it as zero

Could well be the case, I have the Nimbus+ on hand as well but have never looked thoroughly into it's device descriptors.

PR #590, opened 2022-06-18
adds Microsoft SideWinder Precision Pro joystick for macOS & Windows

stevenjamescox:
macOS - `030000005e0400000800000000010000`
Windows - `030000005e0400000800000000000000`

Looks like this:
from_ebaybox_from_ebay

Please lemme know how to change the mappings if need be. They're a little weird and some don't make much sense for in-game usage. Rather, they're an attempt to follow the patterns in the Sega and Arcade mapping guides.

- joystick trigger button: `b0`
  - used as LT

- joystick long vertical button on top of stick left-side: `b1`
  - used as LB
(this button _only_ clicks at its top—it's not like many TV remotes' vertical channel/volume buttons that have both up & down clickability)

- joystick top right-side circle button: `b2`
  - used as RB

- joystick bottom right-side circle button: `b3 ` 
  - used as RT

- "shift" button at the bottom : `b8`
  - used as Start(???) wasn't sure which remaining mapping was best for it
  - this button doesn't create a separate set of potential mappings—that functionality noted on the box was done via ye old SideWinder software

- throttle slider at the moment it reaches its highest possible position: `a3`
  - there is no similar signal received when the slider reaches its lowest possible position
  - the slider's primary gradual-measurement functionality doesn't register in SDL2 Gamepad Tool
  - I thought it best to avoid mapping `a3`(???)

- joystick x-axis: `a0` | joystick y-axis: `a1`
  - used as left gamepad joystick

- joystick twist (z-axis) counter-clockwise: `-a2`|  joystick twist (z-axis) clockwise: `+a2`
  - used as right gamepad joystick (only being able to provide its x-axis, as simply `a2`)

- joystick hat: `dpup:h0.1` | `dpdown:h0.4` | `dpleft:h0.8` | `dpright:h0.2`
  - used as D-Pad

- Button A: `b4` | Button B: `b5` | Button C: `b6` | Button D: `b7`
  - mapped as A/B/X/Y based on their relative positioning

For even more verbose information, [here's this controller's .plist report](https://github.com/gabomdq/SDL_GameControllerDB/files/8934377/Microsoft.SideWinder.Precision.Pro.USB.plist.zip) via [HIDBrowser](https://github.com/OpenEmu/OpenEmu/files/3746691/HIDBrowser.app.zip).


offalynne:
We deliberately omit non-gamepad devices with the exception of fighting-style arcade joysticks. SDL handles these with more generalized joystick support. Thanks for the PR in any case !

PR #598, opened 2022-07-19
Add DragonRise Generic USB Joystick

luizluca:


offalynne:
We deliberately omit maps for this device, see #202

PR #608, opened 2022-08-22
Add Xbox One S Controller mappings

Ev1lbl0w:
Adds two mappings for the Xbox One Controller, one from USB and other from Bluetooth. Tested the mappings on Godot and worked without problems.


offalynne:
Thank you

PR #612, opened 2022-09-01
Added another GUID for TigerGame Playstation Adapter

Zibangarang:
My adapter had a different GUID but it seems identical to the existing entry in every other way. Added another entry with this GUID.


offalynne:
Thank you

PR #615, opened 2022-09-26
Xbox Series controller mapping

Enthuin:
mapping generated from SDL2 2.24.0 controllermap utility


offalynne:
This device should already work as intended without a map, see https://github.com/gabomdq/SDL_GameControllerDB/issues/478

PR #619, opened 2022-10-13
Add Linux mapping for 3rd party PS4 Controller

bonswouar:
Mapping made using gamepad-tool


bonswouar:
@offalynne I've just noticed that in this generated string some keys seem to be duplicated (dpdown, dpleft), any idea if that's normal?

offalynne:
Fixed in 824d1c99fa8ff4c1cdcf0dfb10604fd7149091b9

PR #622, opened 2022-10-21
Correct controller names for portable linux devices

christianhaitian:
Hello,

Not sure what happened from the previous PR I submitted but somehow the names were incorrect.  While it doesn't seem to affect the functionality of games and emulators that make use of the gamecontrollerdb.txt file, I'd like to correct them to match how they are described by the manufacturer in their supplied kernel sources.

Thank you,


offalynne:
@christianhaitian what happened is I cleaned them up! The name field is meant to be readable, so `gameforce_gamepad`, `GO-Advance Gamepad (rev 1.1)`, and `retrogame_joypad` are not acceptable. If you want to clean those fields up I'm glad to merge.

christianhaitian:
I don't understand what that means to clean them up.  They are as the manufacturer describes them in sdl.  For instance, see the SDL description below for the Odroid Go Advance 1.0 units.

image

christianhaitian:
Another example of the description of the device from evtest
image

offalynne:
They need to be **_readable_**. No underscores, no abbreviated revision version, Words in Title Case.

christianhaitian:
I'm not sure why an underscore is not readable to humans as I understand them fine.  Odd choice but ok.  I'll close this PR then.  Thanks for the consideration.

offalynne:
Please see https://github.com/christianhaitian/SDL_GameControllerDB/pull/1, revise if you'd like

christianhaitian:
Doesn't meet the objective I was looking for in matching the name as supplied by the manufacturer.  Do as you'd like with it.

offalynne:
Why are you trying to match "the name as supplied by the manufacturer"

offalynne:
For posterity and the sake of future contributors

>  _Name is the human readable string for the device_
https://wiki.libsdl.org/SDL_GameControllerAddMapping#:~:text=human%20readable%20string%20for%20the%20device

> _Please make sure to check that the name is a good description of the controller._
https://github.com/gabomdq/SDL_GameControllerDB#:~:text=Please%20make%20sure%20to%20check%20that%20the%20name%20is%20a%20good%20description%20of%20the%20controller.

christianhaitian:
Why would I not?  They designed the hardware and provided the kernel sources as intended right down to the description.  I reviewed those links and saw no mention that underscores were an issue but I may have missed tht point.

offalynne:
>  *"They designed the hardware and provided the kernel sources as intended"*

The strings in the examples you provided were automatically populated straight from the HID descriptions and were presumably "good enough" for the manufacturers, since as you say, description has no impact on function. Frequently Chinese manufacturers reuse or repurpose boards as is happening here, so the descriptions were changed to match the actual device the controller is found in.

It's unclear to me why this a problem for you; feel free to articulate why, if you would like.

Issue #609, opened 2022-08-24
New mappings for SDL 2.26


slouken:
This is here just as an FYI, SDL 2.26 is adding hashing of some additional data to help reduce collisions between mappings for difference devices that share the same USB VID/PID. The exact algorithm is still in flux, but once finalized, you should be able to use controllermap from the new SDL to create mappings that can more easily be upstreamed in the future.

offalynne:
@slouken can you extend a hand to GeneralArcade to see if they can update their graphical tool or at least release the source since that’s what the vast majority of our contributors use ?

A heads up when Steam’s mapper gets this change as well would be appreciated.

Ryochan7:
> @slouken would you please considering reaching out to [GeneralArcade](http://www.generalarcade.com/gamepadtool/) to see if they can update their graphical tool or at least release the source since that’s what the vast majority of our contributors use ?
> 
> A heads up when Steam’s mapper gets this change as well would be appreciated.

It isn't going to happen. Tool hasn't even been updated in 4 years. I haven't worked on it in a long time but I got a similar tool that does the job equally well. The AppImage should still work.

https://gitlab.com/ryochan7/sdl2-gamepad-mapper

SDL2_Gamepad_Mapper-0.0.3-x86_64.AppImage
https://drive.google.com/file/d/12CtYjPbP8YLN7dCIJiKX5EcNdf4PIz4g/view?usp=sharing

offalynne:
@Ryochan7 Great for Linux, but the QT project doesn't build on Windows or Mac OS

slouken:
It looks like at least on Windows you can drop in the new SDL2.dll into [GeneralArcade](http://www.generalarcade.com/gamepadtool/) and get the new format GUID and mapping:
```
030024c8a306000022f6000000000000,Cyborg V.3 Rumble Pad,platform:Windows,a:b1,b:b2,x:b0,y:b3,back:b8,guide:b12,start:b9,leftstick:b10,rightstick:b11,leftshoulder:b4,rightshoulder:b5,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,leftx:a0,lefty:a1,rightx:a2,righty:a4,lefttrigger:+a3,righttrigger:-a3,
```

slouken:
> https://gitlab.com/ryochan7/sdl2-gamepad-mapper
> 
[...]

PR #625, opened 2022-10-24
Mayflash N64 Controller Adapter

briaguya-ai:
image

`VID_0079&PID_1879`

I wasn't able to configure the device on linux (jstest didn't recognize the device at all) but i was able to configure and test on windows

There are also adapters that come in the same packaging with different hardware ids `VID/PID: 0x0e8f/0x3013`, those have an overlapping guid with a snes controller adapter that's already in the db, so i'm waiting on https://github.com/gabomdq/SDL_GameControllerDB/issues/609 to add those


offalynne:
Thank you for the PR.

> VID/PID: 0x0e8f/0x3013

I have this on hand. #609 doesn't help, because the name field is the same: `USB GamePad`.
In the past where possible I've ambiguated mappings so both devices work. We can do that here for the desktop mappings (see below). 

The only sticking point is the alpha face buttons -- the usual Nintendo layout dictates we map SDL2 `a` and `b` fields with buttons `B` and `A` respectively (south and east face buttons per Xbox layout), where N64 mappings are `a -> A`, `b -> B`. In order to support both devices layouts adequately, I'll switch the SNES face button fields to match the button labels, making the SNES mapping less technically correct, but adequate for coverage of both devices.

Window, Linux

```
Field            HuiJia SNES    Mayflash N64

a                 b2            b1
b                 b1            b2
x                 b3            --
y                 b0            --
back              b8            --
start             b9            b9
dpup             -a1            h0.4
dpdown           +a1            h0.8
dpleft           -a0            h0.2
dpright          +a0            h0.1
leftshoulder      b6            b6
rightshoulder     b7            b7
lefttrigger       --            b8
leftx             --            a0
lefty             --            a1
rightx            --            a3~
righty            --            a2
~~

Mac

```
Field            HuiJia SNES    Mayflash N64

a                 b4            b4
b                 b2            b2
x                 b6            --
y                 b0            --
back              b16           --
start             b18           b18
dpup             -a1            h0.4
dpdown           +a1            h0.8
dpleft           -a0            h0.2
dpright          +a0            h0.1
leftshoulder      b6            b12
rightshoulder     b7            b14
lefttrigger       --            b16
leftx             --            a0
lefty             --            a2
rightx            --            a6~
righty

briaguya-ai:
>  because the name field is the same

hmm, maybe instead of just using the name as a part of the guid it might make sense to use some other possible differentiating info

for the n64 adapter with `VID/PID: 0x0e8f/0x3013` i see

image

does the snes adapter have the same number of axes/buttons reported in jstest?

offalynne:
I don't know. You'll have to take proposals for SDL changes upstream to SDL.

briaguya-ai:
>  take proposals for SDL changes upstream to SDL.

I completely understand, I just figured knowing if the adapter had any differentiating info could help determine the feasibility of such a proposal.

PR #653, opened 2023-01-05
Support "button label" hint for 8BitDo Pro Wired 2

lbschenkel:
Controllers such as 8BitDo Pro Wired 2 actually have their physical labels according to the "Nintendo layout" (bottom = B, east = A) and not the "Xbox layout" (default, bottom = A, east = B).

SDL supports the SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS for this purpose, which hints to the application that the user wants to follow the labels in the controller, and not the physical layout of the buttons [1]. That hint defaults to 1.

The mappings for this particular controller are made in the "default layout" (hint=0) and do not support the "Nintendo layout" (hint=1), which is inconsistent with how SDL works by default.

This commit addresses that by introducing a new copy of each one of the three mappings for this controller (as it has 3 operations modes which result in 3 different GUIDs [2]), which will activate depending on the hint.

Naturally there are other controllers with the "Nintendo layout" that would benefit from the same improvements, however the author of this commit is restricting the scope of the change to the controller he personally has and can test.

[1] https://github.com/libsdl-org/SDL/issues/6117
[2] https://github.com/gabomdq/SDL_GameControllerDB/pull/650


lbschenkel:
Another thing: I did not test what would happen if old SDL versions (the ones that don't know about the hint in the file) consume this, if the first or last entry wins. A careful sorting of the entries would address that, though (something that could be checked/enforced by a script).

You probably should hold off merging until I can test more. As I said, I just want to hear you out before doing more work.

offalynne:
SDL uses the `SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS` hint, the community db does not :-)

offalynne:
Do feel free to take your mappings including the label hint upstream to the SDL project

lbschenkel:
> SDL uses the `SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS` hint, the community db does not :-)

That is fair, but may I know the reasoning behind that decision — just for curiosity's sake. I'm wonder what it is given that it is totally possible to do it in a way that doesn't break applications using older (pre-hint) SDL. 

> Do feel free to take your mappings including the label hint upstream to the SDL project !

Yes, every submission I'm doing here I will also do upstream as well (I already did for the first one). It's just that you're way faster in approving them and I prefer not to pile up submissions, but do in sequence.

offalynne:
> may I know the reasoning behind that decision

I (and several other significant contributors) do not like it -- it is a maintainability nightmare (having to screen for/add it to submissions), unnecessary duplication of data, bad for back-compat, etc, etc. 

the job of resolving label/layout differences should be handled by gamepad *type* data and has no place in the mapping itself in my (strong) opinion. multiple (solved, to my knowledge) bugs have arisen in SDL from this completely unnecessary complication. evidently SDL project contributors (specifically Valve) felt it was useful -- if they want it for the official db that's fine, our omitting it presents no interoperability issue :-)

lbschenkel:
Just a follow up question, then: this DB should always have the mappings using the positional layout, right? (Basically `SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS=0`.) This is how I contributed mine, I want to make sure I did it right.

offalynne:
> this DB should always have the mappings using the positional layout, right?

Yes, per README

> not sure if this approach would end up working well

8BitD boot modes have a unique VID+PID. It is the intent of 8BitDo (who contribute to this db) that the mapping of each mode match position and not label.

> the type concept would need to be broadened a bit

You’ve identified a potential use for the label hint that SDL itself does not use :-) feel free to take the idea upstream

> makes sense to me why SDL does it at the individual controller mapping

I think we understand each other, but have come to opposite conclusions. Feel free to lobby SDL with any changes you feel appropriate and we may well (or not) inherit them in the future.

lbschenkel:
>> not sure if this approach would end up working well

> 8BitD boot modes have a unique VID+PID. It is the intent of 8BitDo (who contribute to this db) that the mapping of each mode match position and not label.

A clarification: I was commenting about the suggestion of having the label/layout differences being attached to controller types in SDL, instead of mapping. Nothing to do with this PR and/or 8BitDo. Reading it again, it might not have been clear.  

>> makes sense to me why SDL does it at the individual controller mapping

> I think we understand each other, but have come to opposite conclusions. Feel free to lobby SDL with any changes you feel appropriate and we may well (or not) inherit them in the future.

Sure, this was just an off-topic chat — not an attempt to persuade you to merge the PR. Will leave this alone now as there are better forums for this kind of talk.

offalynne:
Thanks for clarifying. I definitely consider this open to discussion and am not trying to shut out potential changes, please feel free in the future to open an issue for further discussion if you want, may also lead to additional input from other parties.

PR #655, opened 2023-01-16
Add another Xbox Series X controller mapping for Windows

RT222:
I was wondering why the share button wasn't recognized on my Xbox Series X controller, and I noticed that the guid was different from the Xbox Series controller that is currently specified for Windows in the database, so I added it to the list.


offalynne:
XInput devices are meant to be handled by SDL's driver on Windows and should not require a mapping. If the device is mapped incorrectly you should take the issue [upstream to the SDL project](https://github.com/libsdl-org/SDL/issues).

offalynne:
Out of curiosity, could you check what you get from the [old mapping tool](http://www.generalarcade.com/gamepadtool/) ?

RT222:
Sure, here's the output : 
```
SDL2 Gamepad Tool v1.2 by General Arcade (compiled with SDL version 2.0.7, DLL version 2.0.7)
Website: http://generealarcade.com/gamepadtool/

Searching gamepads...
Found 1 gamepad(s):
"XInput Controller", 030000005e040000120b000000007801 (mapping available)
Environment variable "SDL_GAMECONTROLLERCONFIG" is not defined
Checking if new mappings available from github: https://github.com/gabomdq/SDL_GameControllerDB
Mapping string for "XInput Controller" is "00000000000000000000000000000000,XInput Controller,a:b0,b:b1,back:b6,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b10,leftshoulder:b4,leftstick:b8,lefttrigger:a2,leftx:a0,lefty:a1,rightshoulder:b5,rightstick:b9,righttrigger:a5,rightx:a3,righty:a4,start:b7,x:b2,y:b3,platform:Windows,"
```
The misc1 button is not defined for XInput Controllers. I tried to add it manually to the gamecontrollerdb.txt file, but it didn't work. It works only if I add a new entry with the guid of my controller and the misc1 button defined.

PR #659, opened 2023-02-05
added elecom 8button gamepad on linux

planetis-m:


offalynne:
GUID is a duplicate of https://github.com/gabomdq/SDL_GameControllerDB/blob/fabf890296b587cb4099034b4b4a7a974688eb17/gamecontrollerdb.txt#L1413 (see https://github.com/gabomdq/SDL_GameControllerDB/pull/654) and since we're not using the 2.26 CRC field (see https://github.com/gabomdq/SDL_GameControllerDB/issues/609) I won't be merging this. Feel free to take the mapping upstream to the SDL project, it should be approved there.

planetis-m:
Thanks for the info! I was trying to figure out why it doesn't work and that was helpful.

offalynne:
If you're able & willing I would like to ask you for some follow up info on the device, could you reach me by email on alviscera@gmail.com ? Either way, thanks for submitting and reading.

PR #665, opened 2023-02-23
PDP XBox Series X

gcongedo:
Added mapping for PDP XBox Series X

Closes https://github.com/MrCrayfish/Controllable/issues/359


offalynne:
Please use http://www.generalarcade.com/gamepadtool/ to generate a mapping; Antimicro is not suitable for this task.

PR #667, opened 2023-02-26
add new stadia controller mappings

greyltc:
mappings made with Arch Linux, sdl2 2.26.3-1 and sdl2-gamepad-mapper 0.0.4-3 (commit 7d3f7ad9780933bbeaa8bb76f5db3d501e5c1e2d) for my stadia controller connected both via USB and Bluetooth

I don't know why my IDs disagree with the existing ones for the stadia controller.


offalynne:
Looks like the only difference is the CRC field (which is meant to be removed), otherwise these match the exiting mappings, so no changes needed! Thanks.

PR #671, opened 2023-03-11
Update mappings of gamepads affected by recent SDL change.

divVerent:
This adds updated mappings for for changes in
https://github.com/libsdl-org/SDL/commit/6f1f586086a18ab32ddccc05e4616aac02774a1b and https://github.com/libsdl-org/SDL/commit/de3909a190f6e1a3f11776ce42927f99b0381675

I have made a change to SDL to treat default Android mappings closer to reality and documentation on Android, by treating gamepads in which all of X, Y, Z, RX, RY, RZ axes exist to to return the axes in the order X, Y, RX, RY, Z, RZ.

See e.g.
https://developer.android.com/develop/ui/views/touch-and-input/game-controllers/controller-input which documents Z and RZ to be for the right stick (and thus belonging together); also, a whole lot of Xbox-ish gamepads export the right stick as RX, RY opposing Google's documentation (but matching Linux's at https://www.kernel.org/doc/html/latest/input/gamepad.html) and the triggers as Z, RZ, which SDL's automatic mapping now handles.

This requires the changes to the mappings included in this commit - which adds them as extra mappings to not break with previous SDL versions. Note that it is quite possible that some of these gamepad entries could even be removed now; I cannot say for sure if any of the added mappings are optional or required. In particular some of the Xbox ones may be redundant with defaults but still differ from default by _not_ mapping some buttons or axes.


offalynne:
I'm wary of version-specific breaking changes like this as
1. We try not to be too prescriptive of SDL version
2. Even though it says SDL on the tin, many (large) projects that don't use SDL but borrowed some of it's code at one point rely on these mappings

Would you consider amending your upstream patch to flip a GUID bit somewhere so that the necessary changes can coexist with older mpings, and future mapping submissions will be identifiable to prevent collision ?

Obviously the best case is your future-forward proposal, but the reality gets messy.

divVerent:
Sent https://github.com/libsdl-org/SDL/pull/7444 to SDL - if that goes through, I'll update this PR accordingly - basically, restoring the previous line for every change, but changing the new one to have an 80 as the last SDLID byte.

offalynne:
Perfect, thank you for this !

offalynne:
@divVerent since Sam has OK'd that change, I'll close this PR and you can submit another for any new mappings that aren't correctly covered by the HIDAPI driver (though if I understand your changes correctly, no new mappings should be required?) unless you would rather keep your changes on this branch

divVerent:
I'd prefer updating this one instead; no mappings seem entirely redundant, but some of the mappings MAY be unnecessary (they differ from default only by NOT mapping some buttons or axes that exist).

divVerent:
Will need squashing and rewording, though.

Will this be squashed to a single commit, or merged? If squashed, I'll edit the first post here and that should do, I guess?

divVerent:
Comment adjusted - so if you squash, all should be fine.

divVerent:
FYI after these changes, the following are exact clones of SDL's default layout for Android:

```
05000000b404000011240000dfff3f00,Flydigi Vader 2
050000004c05000068020000dfff3f00,PS3 Controller
050000004c050000c405000000783f00,PS4 Controller
050000004c050000c4050000ffff3f00,PS4 Controller
050000004c050000cc090000fffe3f00,PS4 Controller
050000004c050000cc090000ffff3f00,PS4 Controller
050000004c050000e60c0000ffff3f00,PS5 Controller
05000000f8270000bf0b0000ffff3f00,Razer Kishi
050000003215000005070000ffff3f00,Razer Raiju Mobile
050000003215000007070000ffff3f00,Razer Raiju Mobile
050000005e0400008e02000000783f00,Xbox 360 Controller
050000005e040000fd020000ffff3f00,Xbox One Controller
050000005e040000000b000000783f80,Xbox One Elite 2 Controller
050000005e040000e002000000783f00,Xbox One S Controller
050000005e040000ea02000000783f00,Xbox One S Controller
050000005e040000fd020000ff7f3f00,Xbox One S Controller
050000005e040000120b000000783f80,Xbox Series Controller
```

Some of these become redundant with this change (and thus end with 3f00, not 3f80), but most are pre-existing.

However, may still be worth keeping them as they help games that do NOT have a default mapping like SDL does. Especially as most of these sound like popular controllers.

offalynne:
Looks good, thanks

PR #674, opened 2023-03-28
Added PowerA Xbox One with GUID 030000005e040000ff02000000007801

mghoffmann:
Adding this controller to work with MonoGame. See https://github.com/MonoGame/MonoGame/issues/6263


offalynne:
See https://github.com/MonoGame/MonoGame/issues/6263#issuecomment-1487078099

PR #675, opened 2023-04-02
Mega World Logic 3 mapping

robdobes:
Another  Mega World Logic 3  controller . Just a different guid and swapped rightx righty
linux os
lsusb shows
 ID 07b5:004f Mega World International, Ltd USB Game Controller

on the front of the controller it says Logic3
on the back it says Model No  JP279   Crusader  Logic 3 International ltd (UK)


offalynne:
Thank you

PR #682, opened 2023-04-29
Microsoft Sidewinder Force Feedback 2 Joystick

lordloki:
Added Microsoft Sidewinder Force Feedback 2 Joystick on Windows


offalynne:
the gamepad mapping db deliberately omits flightsticks see https://wiki.libsdl.org/SDL2/CategoryJoystick

PR #684, opened 2023-05-05
Update gamecontrollerdb.txt

RyadMax:
Added: Relax USB Controller.


offalynne:
We omit this GUID, see https://github.com/gabomdq/SDL_GameControllerDB/issues/202

PR #685, opened 2023-05-14
Added Xbox Series Controller

Parrawk:
Apparently this GUID is missing from the list. It was obtained with GLFW 3.3.8 `glfwGetJoystickGUID`.

I have 3 Xbox Series controllers and this is the only one that wasn't being recognized in Minecraft with the MidnightControls mod. The controller has the latest  firmware and installing `xpad-neo` and `xone` didn't solve the issue.

After reading a similar problem in PR #668 and running Gamepad Tool, Gamepad Mapper and GLFW I noticed that they were reporting different GUIDs (Gamepad Tool and GLFW were the same), so I'm adding the GUID from Gamepad Tool/GLFW like in #668 and the rearranged mapping from Gamepad Tool since it's identical to the ones above it. I also changed the name from `Xbox Series X Controller` to just `Xbox Series Controller`.


offalynne:
Thank you. Can you get a mapping for the Share button ? You can make a new mapping and skip everything but the one button to find the value, or use a tool like jstest-gtk. It should be mapped to the "misc1" key.

Parrawk:
Got the mapping by the just pressing the Share button and skipping everything else. Thanks.

Issue #664, opened 2023-02-17
New Xbox series X mapping for Linux after update


PKGaspi:
After updating my Xbox series X controller's firmware, the mapping is broken for Linux. This is the new mapping:

```
030000005e040000120b00000f050000,Microsoft Xbox Series S|X Controller,a:b0,b:b1,y:b3,x:b2,start:b7,back:b6,leftstick:b9,rightstick:b10,leftshoulder:b4,rightshoulder:b5,dpup:b12,dpleft:b14,dpdown:b13,dpright:b15,leftx:a0,lefty:a1,rightx:a3,righty:a4,lefttrigger:a2,righttrigger:a5,platform:Linux
```

offalynne:
7d416272d904e2d53eed353abda7468ba4fd023c

2000Slash:
Hey I have a quick question about this. (I don't really know if I should open a new issue). My Xbox Series X Controller with the exact same guid and the newest version still uses the hat input for the dpad instead of buttons, just like the other Xbox Series Controllers.
```
dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1
```
instead of
```
dpdown:b13,dpleft:b14,dpright:b15,dpup:b12
```
Since the entry for this controller was added, the dpad does not work for me in some programs (pcsx2). I use the xpad driver and face this issue on my main arch system and a fresh ubuntu install.

2000Slash:
Just tried it, but since xpad is the default driver for arch and I don't have another, it just doesn't get recognized anymore.

offalynne:
Can you try with the *"fresh ubuntu install"*, please ?

offalynne:
@PKGaspi can you provide any insight as to what your setup is like to compare? What distro you're on, what driver might be handling the Xbox pad, etc?

2000Slash:
I just tried in on a fresh ubuntu live and it is exactly the same as on my arch machine. The driver seems to be xpad. When I try to run pcsx2 the input is only accepted when I use the hat mappings. When I unload xpad the controller is not recognized anymore

offalynne:
Last thing, can you please confirm either with an Xbox Console or on Windows with the Xbox Accessories app that your controller is on the latest firmware? If so please open a PR with the relevant change.

2000Slash:
Yup just tried to update it on my windows pc, but it says its on the newest version.

PKGaspi:
[...]

PR #686, opened 2023-05-22
Add Xbox Series controller for Linux

micke1m:
Generated with SDL2 controllermap


offalynne:
For (frustrating) reasons I won't bother getting into the actual final GUID for the purpose of this DB from your mapping is `030000005e040000120b00000f050000`.

See https://github.com/gabomdq/SDL_GameControllerDB/issues/664 re. dpad discrepancies: feel free to comment there if you have further findings or any input that would motivate a change from current db values.

PR #688, opened 2023-06-04
Fixed the SDL tool name, and added a link to its source code

denilsonsa:
I previously had trouble trying to find the official SDL tool. With this commit, I'm using the correct name for the tool, as well as providing a direct link to the tool in the SDL repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,8 +36,8 @@ Third party cross-platform tool with GUI (Windows, macOS and Linux)
 ### [SDL2 Gamepad Mapper](https://gitlab.com/ryochan7/sdl2-gamepad-mapper/-/releases)
 Open source GUI app for authoring mappings. Builds available for Windows and Linux.
 
-### [SDL2 ControllerMap](https://www.libsdl.org/download-2.0.php)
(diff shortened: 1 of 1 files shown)


artcfox:
@offalynne 

I think this change should to be reverted. I recently had to use the game controller mapping tool, and was confused because it is not called gamepadmap for SDL2, that is the name of the SDL3 tool.

The original instructions were correct for SDL2, but since the main branch now tracks SDL3, you first need to switch to the SDL2 branch in order to find the controllermap tool: https://github.com/libsdl-org/SDL/blob/SDL2/test/controllermap.c

```
git clone https://github.com/libsdl-org/SDL
cd SDL
git checkout SDL2
cd test
./configure
make
./controllermap
```

denilsonsa:
I wasn't aware the main branch was already for SDL 3. In such case, it might be worth keeping both the old name and the new name, for people who have either the version 2 or the version 3. In any case, the tool name should contain a link pointing to where the tool can be obtained.

offalynne:
Feel free to open a PR with those changes

offalynne:
https://github.com/gabomdq/SDL_GameControllerDB/commit/8f2ddc0de02dfe4bd6b0b6cb22019153267b95dc

PR #690, opened 2023-06-12
Add X-box One Elite 2 with version 0x407

iacore:
```
# evtest

Input driver version is 1.0.1
Input device ID: bus 0x3 vendor 0x45e product 0xb00 version 0x407
Input device name: "Microsoft X-Box One Elite 2 pad"


# lsusb -v

Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               2.00
  bDeviceClass          255 Vendor Specific Class
  bDeviceSubClass        71 
  bDeviceProtocol       208 
  bMaxPacketSize0        64
  idVendor           0x045e Microsoft Corp.
  idProduct          0x0b00 
  bcdDevice            4.07
  iManufacturer           1 Microsoft
  iProduct                2 Controller
```

There are also 4 buttons at the back labeled P1 P2 P3 P4. I named them `paddle1` and etc.

I also updated the controller with version 0x408. It seems to be the same controller.

Is there a way for SDL to detect controller but ignore the USB device version (`bcdDevice`)? It's not useful in this case.


offalynne:
Paddles are a little weird: https://github.com/libsdl-org/SDL/blob/dec0dbff13d4091035209016eb2d0dd82c9aba58/include/SDL3/SDL_gamepad.h#L101-L104

image

can you confirm the mapping is correct per the above ?

iacore:
sorry it's wrong. I thought simply paddle1 is P1 and so on

PR #693, opened 2023-06-17
Update gamecontrollerdb.txt

AetheralMeowstic:
Adds a Rock Candy Wired Controller


offalynne:
Which platform did you actually test this on? The GUID and mappings are certainly not the same on all 5 platforms.

offalynne:
I assumed based on GUID the mapping was authored for Linux, in which case it is a duplicate of 7f22bae4d3f6a76dd6405156fcac14761373f664

PR #694, opened 2023-07-09
Add DaemonBite SNES/NES controller USB adapter

danboid:


offalynne:
thanks! does the mapping match the readme (eg A->B X->Y etc)?

offalynne:
Omitting this since it's based on a generic and configurable microcontroller

PR #702, opened 2023-09-03
Add DualSense Wireless Controller

FrogTheFrog:
This is the PS5 Controller "alternative" you get when you compile the SDL with `libudev`.

I have omitted the `crc:5657` field.


offalynne:
The `6800` suffix indicates this device is being handled by SDL's hidapi driver. Does the device already work as intended in SDL without this mapping ?

FrogTheFrog:
> The `6800` suffix indicates this device is being handled by SDL's hidapi driver. Does the device already work as intended in SDL without this mapping ?

Yes. Also, this mapping was generated by SDL2. However, since I needed to customize it for my purposes and I didn't find it in the files, I decided to create a PR.

If it is redundant and not needed, please close the PR :)

PR #704, opened 2023-09-13
Add mapping for GPD Win Max 2 (6800U model) controller

frikisama:
A new mapping is needed for the built in controller in the GPD Win Max 2 (6800U model, from 2022). It has a mostly standard XBox layout, but presents itself in a weird way, apparently.

This mapping was extracted from the Steam config.vdf and successfully tested with Controlify (Minecraft plugin), which would previously just recognize an unknown, unmapped joystick device.

030000005e0400008e02000001010000,Atari Xbox 360 Game Controller,a:b0,b:b1,back:b6,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b8,leftshoulder:b4,leftstick:b9,lefttrigger:a2,leftx:a0,lefty:a1,rightshoulder:b5,rightstick:b10,righttrigger:a5,rightx:a3,righty:a4,start:b7,x:b2,y:b3,platform:Linux


offalynne:
Thank you!

PR #711, opened 2023-10-08
Add Retrode USB gamepad adapter

revvv:


offalynne:
Thank you ! Does this device support 6b MD gamepads as well ?

revvv:
Yes, I got cheap clones and they work!

offalynne:
Could you please author mappings using the Sega layout here: https://github.com/gabomdq/SDL_GameControllerDB/blob/master/mapping_guide.png

and post in this thread, so we can decide how to handle the superset, assuming there's no difference on the PC's end wrt GUID ?

revvv:
My submitted configuration is meant for the SNES pads, but with A/B+X/Y flipped.
As the 4 controllers seem all identical to SDL2 you can't provide a different configuration for the MD pads.
6 button MD pads now get:

    first row buttons:  X Y Z -> L Y R
    second row buttons: A B C -> X A B

At least this is how my cheap MD clone (AFUNTA) behaves. (Has 6 buttons and a start button)
The only OS which allows a separate configuration is Linux and RA might be able to make use of it. (Not tried.)

My submission maps also the left stick. Maybe this is against your standards. Please let me know.
But in the games I play I prefer playing the left stick and not the DPAD, which would also work. So you have a choice.

offalynne:
> _6 button MD pads now get..._

In order to minimize confusion here, ignoring everything else, please run thru Win/Mac/Linux using the 6 button MD pads and create a *new SDL mapping string* to post here that matches the Sega format: https://github.com/gabomdq/SDL_GameControllerDB/blob/master/mapping_guide.png

I will compare that mappings to the ones in your PR (assuming you've correctly followed the Nintendo A -> SDL b, Nintendo A -> SDL b format per the above image) and decide how to best handle a mapping that makes best-sense for both.

Thank you !

revvv:
Wow, thank you very much for all your effort!

PR #713, opened 2023-10-11
Add Xiaomi Black Shark Gamepad 2.0 (L)

eVenent:
This adds a Windows mapping for the Xiaomi Black Shark Gamepad 2.0 (L).

Default mapping is weird and d-pad works as face buttons. This config fixes d-pad and shoulder buttons to work as intended.


offalynne:
No mappings for the right hand controller ?

eVenent:
They are selling them separately and usually only the left one. The right one with the touchpad is not so easy to get, but when I find it and buy, I will definitely send a second commit.

PR #715, opened 2023-11-01
Updated link to SDL testcontroller and added entry for AntiMicroX

slouken:


offalynne:
Any reason in particular for adding AntiMicro ? I believe the SDL version is out of date and I don't think we need to link two of Travis' projects for the same purpose.

slouken:
Mostly because it's the one I've heard people talk about most when discussing controller mapping software. Feel free to disregard that portion of the patch.

PR #719, opened 2023-11-24
Add SDL3 function name

Susko3:
See https://github.com/libsdl-org/SDL/blob/f40f272107ff9e8cfe96f3a531944e7cc411cee6/test/testcontroller.c#L1945

Also adds C syntax highlighting to make it look nicer.


offalynne:
Thanks 👍

PR #726, opened 2023-12-19
add 8bitdo n64 modkit (linux)

briaguya-ai:
actually make a PR out of https://github.com/gabomdq/SDL_GameControllerDB/pull/724


offalynne:
This is a duplicate of the map immediately above it. Both have the left trigger mapped so I have no idea what #724 was about. The only difference in the mapping is `back:b12`, but there's no such button on the controller or modkit ??

briaguya-ai:
i'm not sure, it's not an exact duplicate because the ids are different `05000000c82d00006928000000010000` vs `050095fcc82d00006928000000010000`

as for back, maybe it's one of these?
image

@Vulto can you provide more context on this?

offalynne:
> the ids are different

the IDs are the same. the difference is the CRC which is meant to be removed from the GUID.

> as for back, maybe it's one of these?

it is not. i have the device in hand.

Vulto:
> i'm not sure, it's not an exact duplicate because the ids are different `05000000c82d00006928000000010000` vs `050095fcc82d00006928000000010000`
> 
> as for back, maybe it's one of these?
> image
> 
> @Vulto can you provide more context on this?

Ok.
Yeah is this one of the picture.
The controller works fine after i remap the Z button. It's no mapped correctly out of the box.

offalynne:
Your mapping matches the one in the db and the GUID in your PR was formatted incorrectly. Whatever context you were testing in must not have had the latest db file.

briaguya-ai:
> > the ids are different
> 
> the IDs are the same. the difference is the CRC which is meant to be removed from the GUID.
> 
> > as for back, maybe it's one of these?
> 
> it is not. i have the device in hand.

Got it. Sorry for the trouble, I'll make sure to check/clear CRCs before searching for GUIDs in the DB when PRing mappings users share with me in the future.

retropieuser:
Is 05000000c82d00006928000000010000 correct? In linux any applications I'm attempting to use with the controller the A and B are always the wrong way around I think it should be this, can someone else test this and confirm?

05000000c82d00006928000000010000 ,8BitDo N64,a:b1,b:b0,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b2,leftshoulder:b4,lefttrigger:b6,leftx:a0,lefty:a1,rightshoulder:b5,righttrigger:b10,rightx:a2,righty:a3,start:b9,platform:Linux,

PR #729, opened 2023-12-29
Add ClockworkPi DevTerm for Linux

xordspar0:
This supersedes https://github.com/libsdl-org/SDL/pull/8359, which hasn't gotten any traction.

This is the controller built into the [ClockworkPi DevTerm](https://www.clockworkpi.com/home-devterm). I used test/controllermap.c from SDL 2.28.4 to generate the mappings, then I cleaned it up slightly to match other entries.

photo of the DevTerm, showing the controller


offalynne:
Thanks!

Issue #200, opened 2018-03-06
NYKO AIRFLO EX corrected mapping (SDL 2.0.7, Windows)


eddieray:
030000004b120000014d000000000000,NYKO AIRFLO EX,platform:Windows,a:b0,b:b1,x:b2,y:b3,back:b8,guide:b10,start:b9,leftstick:b11,rightstick:b12,leftshoulder:b4,rightshoulder:b5,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,leftx:a0,lefty:a1,rightx:a3,righty:a2,lefttrigger:b6,righttrigger:b7,

p-groarke:
Please open a PR with the fix, I will gladly accept it.

AliceLR:
I have confirmation from a friend with a Nyko Airflo Ex that this is the correct mapping. However, it looks like the current very broken mapping [comes from SDL or it was sourced from here](https://hg.libsdl.org/SDL/file/c005c49beaa9/src/joystick/SDL_gamecontrollerdb.h).

p-groarke:
You'll have to convince the SDL project itself, as we don't modify upstream mappings ever.

So if you want to fix it, you'll need to send a pr to the SDL project (for the header you linked). Once it is accepted and merged, we will get the fix at the next SDL release.

Cheers

AliceLR:
I submitted [a patch](https://bugzilla.libsdl.org/show_bug.cgi?id=4534) to SDL.

AliceLR:
My upstream patch for this issue (which also adds Linux and Mac mappings) has been open for over a year without any update. Is anyone aware of a better way to submit a patch to SDL? The Nyko Airflo is a pretty reliable controller and it'd be great if its Windows mapping could be fixed.

offalynne:
you might give sam latinga, ryan gordon, or one of the other active developers a ping on the sdl discourse forum

> we don't modify upstream mappings ever
this seems to have relaxed significantly with several merged PRs to this DB doing exactly this, so perhaps you might just PR here and get the maps right in the community DB, at least

offalynne:
@AliceLR https://twitter.com/icculus/status/1285399383505358850?s=21 😺

PR #730, opened 2024-01-01
Update NYKO Airflo EX mappings to match upstream.

AliceLR:
These were fixed upstream 3 years ago (#200) but I forgot to submit the corrections here. Sorry! :(

The upstream macOS mapping seems to have found its way here separately so I didn't change it.


offalynne:
Thanks!

PR #731, opened 2024-01-07
Added missing mapping for triggers on Logitech F310

IovlevYuliy:
The mapping for triggers were absent for Logitech F310 in Windows section.
Tested on two gamepads.


offalynne:
Looks like this needs a bit of workshopping.
`lefttrigger:a2` indicates the full range of -1 to 1 of axis 2 is for left trigger
`righttrigger:-a2` indicates half the range (-1 to 0) of axis 2 is for right trigger
this doesn't make sense as there is then overlap between left and right and they'll both trigger for -1 to 0. it's possible what you want is `lefttrigger:+a2,righttrigger:-a2`, though that also seems a bit weird. This situation resembles the conflict between trigger values when XInput devices are accessed over DInput where there's no real way to tell the trigger input's apart, since they both act on the same axis.

If you can test a bit more and clarify what you're seeing, please do.

IovlevYuliy:
Thank you for your clarification!
You are right, the gamepad was in XInput mode and there is a toggle for switching it to DInput mode.
After that I get independent axes for both triggers, but anyway mapping for triggers are missing, I updated PR with proper mapping.
```
lefttrigger:a2
righttrigger:a5
```

PR #733, opened 2024-01-16
Add Arduino-based SNES to USB gamepad by UltimateMister.

vanfanel:
Added support for Arduino Leonardo based SNES to USB adapter by UltimateMister (https://ultimatemister.com/product/ultimate-deamon-x3-player-snes-sfc) which is designed for the MiSTer FPGA platform but works perfectly fine on GNU/Linux and other OSes.


offalynne:
The Leonardo is an extremely common microcontroller and prescribing a mapping to it without an explicitly reassigned vendor and product ID would harm more than help

PR #734, opened 2024-01-18
Update gamecontrollerdb.txt

retropieuser:
Editing 8Bitdo M30 controller for linux so that Minus and Favourites buttons are mappable

image


offalynne:
What firmware are you on ?

offalynne:
I'm rejecting this PR as
1. The mappings have been copied across for multiple devices, one of which (8Bitdo M30 modkit) does not match the model depicted
2. The stated intent is not accomplished as the back and guide button mappings have been nonsensically moved to other obviously-incorrect keys
3. The shoulder and trigger mappings may well reflect valid changes, but given the oversights of points 1 and 2 and that this change is not cited as the PR's intent, I doubt this is the case

PR #744, opened 2024-02-23
Add Wireless GB/GBA Adapter

meet-vlad:
Add Wireless GB/GBA Adapter for Windows


offalynne:
This GUID is from a generic Arduino board and so should not be mapped. Will need to remove other devices that match on this GUID as well, thanks!

PR #748, opened 2024-02-27
Add mapping for GUID returned by XBox Wireless Controller using xpadneo.

karnkaul:
I'm using [xpadneo](https://github.com/atar-axis/xpadneo) for connecting my XBox controller via Bluetooth (on Linux 6.7.4-2), which reports a slightly different GUID than the one listed in the mappings. 

GLFW fails to [find a mapping](https://github.com/glfw/glfw/blob/7b6aead9fb88b3623e3b3725ebb42670cbe4c579/src/input.c#L70-L81) for such a gamepad, despite it working flawlessly when connected via USB (or using [xow](https://github.com/medusalix/xow) with the XBox Bluetooth dongle).

This PR adds another entry with the GUID that xpadneo ends up reporting. GLFW detects the joystick as an XBox Wireless Controller correctly after having it load the modified `gamecontrollerdb.txt` at runtime. Merging this change upstream will enable GLFW users to [embed the mappings at build-time](https://github.com/glfw/glfw/blob/7b6aead9fb88b3623e3b3725ebb42670cbe4c579/CMake/GenerateMappings.cmake#L4) and not need to package/load the text file at runtime.


offalynne:
Thank you

PR #750, opened 2024-04-07
Fixed the Atari Classic joystick button mappings

ellisvelo:
This change also aligns the mappings with SDL 2.30.x:

https://github.com/libsdl-org/SDL/blob/release-2.30.x/src/joystick/SDL_gamecontrollerdb.h#L595.

I also changed the Windows mapping to have the same name as the Linux Atari Classic Controller mapping.


ellisvelo:
@offalynne, thanks for fixing the name of the controller so it's more consistent, but the mapping is incorrect for me. The PR had made the settings the same as what SDL has in their source code. Are you seeing something different?

https://github.com/libsdl-org/SDL/blob/release-2.30.x/src/joystick/SDL_gamecontrollerdb.h#L595

offalynne:
Can you elaborate on how the mapping is wrong, exactly ? I worked closely with a contributor to make sure the Pro/Classic pad mappings in the DB are consistent and sensible. What is mapped incorrectly, exactly ?

ellisvelo:
Sure thing, the mapping for this repo is `b:b1`, but the one in SDL2 is `x:b1`. The following link will take you to the definition that SDL has: https://github.com/libsdl-org/SDL/blob/release-2.30.x/src/joystick/SDL_gamecontrollerdb.h#L595

 I only have the one Atari VCS Classic controller with id `03000000503200000110000000000000` so I don't know if the behavior is different or not with the other id's.

offalynne:
OK, what I'm asking is *why* the X mapping is preferable to B. The B mapping gives a result consistent with the "Modern" controller, and behaves according to other software on the VCS itself. I see no reason to map the classic button to X instead of B. The Guide/Back mapping changes do seems sound; if those are consistent I will make that change.

ellisvelo:
When I was testing, I noticed that the lower button on the classic controller wasn't working for me. I copied the mapping from the upstream mapping and that worked and I didn't investigate any further. The Guide/Back buttons worked for me without any problems.

Thanks for looking into this and supporting the game controller db. 👍

offalynne:
Here's the SDL change for those mappings
https://github.com/libsdl-org/SDL/commit/89fd9821dea5c7b8cbd043efd01bc68156b378ee

From my testing with another user, we found the mappings currently in this repo both correct, and consistent with the "Modern" controller. If a conflict between SDL's db and this db is causing an issue, either we need to PR our mapping into SDL, or confirm our mapping is wrong in order to conform to SDL's.

ellisvelo:
@offalynne , you might find this interesting.  Here is the link for the Atari vcs repo where the x mapping was most likely introduced: https://github.com/atari-vcs/libsdl2/blob/atari%2Fvcs/src/joystick/SDL_gamecontrollerdb.h#L557

offalynne:
The mapping was introduced in the change I linked above. I've asked a colleague to look into this, once they have hardware in-hand and can confirm the mapping across platforms the necessary changes can be made to this repo and upstream SDL, if necessary.

PR #751, opened 2024-04-23
Add mappings for `1105` and `1405` Xbox Series X Controllers on Linux

akien-mga:
This is for a first-party Xbox Series X Controller plugged with USB. There are already mappings for the bluetooth UID it seems.

Draft for now as I believe this controller (at least with the current firmware) needs a mapping for macOS too (wireless mode).
I'll test on Windows tomorrow to see if it needs a dedicated mapping too.

Mapping generated with SDL2's `controllermap`.


offalynne:
MacOS does not support Xbox pads over USB. XInput, WGI and/or GDK GameInput handle USB support on Windows. No further mapping should be necessary.

akien-mga:
I tested on Windows and the same controller works fine in both USB and wireless modes.

I do believe there are issues with macOS wireless mappings for that GUID, at least my team mates had that issue in Godot during a weekend jam. I don't have direct access to a Mac now to test myself, I'll see if I can find one to borrow for a quick test.

akien-mga:
So I updated the firmware for one of my Xbox controllers, and it got a new GUID again, which requires the same mapping.

With my commit there's now 3 times the same mapping for USB Xbox Series X controllers on Linux, with slightly different GUIDs (only the last byte is changing).

Should SDL-like gamepad implementations simply ignore that last byte and use whichever mapping matches the `030000005e040000120b0000` part? If so this might be something we need to fix in Godot more than here.

Xbox controller firmware updates seem frequent so keep track of all of them seems a bit tricky.

offalynne:
Interesting questions but not relevant here. If the device works as-is in controllermap, no mapping is necessary. If not, a mapping is necessary. Changes to ignore particular components of the GUID to normalize support on the end of SDL (or Godot, or whatever) are worth pursuing upstream in those projects; the db is just data. Regardless, we accept mappingings if the GUID-mapping pair is unique as it will improve support for the device with the plethora of existing software that already support ingest of the db.

offalynne:
Merging, feel free to open another PR with Mac mappings

PR #752, opened 2024-04-30
Add/Update mappings for 8BitDo Ultimate 2.4G Controller on Linux (xpadneo) (X-Input and D-Input modes)

xandayn:
Tested with GLFW, the D-Input mappings already existed but lacked paddle support.

In X-Input mode the controller has a different GUID, so I've added it. Unfortunately the paddles don't appear to function in X-Input mode.


offalynne:
Thank you !

PR #755, opened 2024-05-12
Added AYN Odin 2

harmsc12:
This is my first time using Github.  I hope I'm doing all this correctly.  I couldn't figure out how to get the tool, so instead I copied and pasted one of the lines into Godot and changed what I needed to get my Odin working correctly.  Someone else with the same device is welcome to check my work.  Also, there seems to be an issue with my Odin when releasing the L and R triggers.  After the first press, they only go down to 0.5 rather than 0.  If someone knows how to add a recalibration into the line, please do.


offalynne:
Good work, thank you !
> _After the first press, they only go down to 0.5 rather than 0_
Can you try `lefttrigger:+a5` and `righttrigger:+a4` and see if you get a better result ?

harmsc12:
The ABXY match on position thing was what sent me down this rabbit hole in the first place.  As for adding the + signs, I just gave that a try and it fixed the 0.5 problem.  So, do I retract this pull request and issue a new one with the fixed triggers?

PR #756, opened 2024-05-13
AYN Odin 2

harmsc12:
I've added a line for AYN Odin, but there is one caveat that I'm not sure about.

I got the GUID by writing up a quick and dirty tool in Godot.  Basically, I have a label named PadGUID and this line of code:

`$PadGUID.set_text (Input.get_joy_guid(0)) #Device GUID`

The number it returned is shorter than all the ones in the database.  Here it is next to the number for my Switch controller, for comparison:

```
Odin 2:               4f64696e20436f6e74726f6c6c6572
My switch controller: 030000006f0e00008401000011010000
```

Interestingly, the number for my Odin works just fine when I feed it back into Godot with this statement:

`Input.add_joy_mapping("4f64696e20436f6e74726f6c6c6572,AYN Odin 2,a:b1,b:b0,back:b4,dpdown:b12,dpleft:b14,dpright:b13,dpup:b11,leftshoulder:b9,leftstick:b7,lefttrigger:+a5,leftx:a0,lefty:a1,rightshoulder:b10,rightstick:b8,righttrigger:+a4,rightx:a2,righty:a3,start:b6,x:b3,y:b2,platform:Android,", true)`

I had a slightly different version of that line, minus the Godot stuff, in an earlier, now-retracted pull request, and when I went to add in a small fix, that was when I noticed the discrepancy.  I don't want to accidentally break something with an incorrectly-formatted GUID, so I'm making sure to mention all this.  I don't know if it's an obscure bug somewhere or just the Odin devs being weird.

Things I made sure to do with this mapping:

ABXY buttons mapped by position.
Correct mapping of the left and right triggers.
Correct the minimum value on the left and right triggers.
Test the mapping in a Godot APK to make sure everything works correctly.


harmsc12:
Also, you'll see a different GUID in the middle commit.  That was from a different tool, and using it in the Godot line didn't work.  As far as Godot is concerned, the shorter GUID seems to be the correct one.

offalynne:
Can you try right-padding the GUID with 0s and see if Godot still picks up the mapping ?

harmsc12:
> Can you try right-padding the GUID with 0s so it reaches a normal length and see if Godot still picks up the mapping ?

I just checked that and compiled a new APK.  That broke the mapping.  For whatever reason, the folks over at AYN gave this device a shorter GUID.

offalynne:
Nothing to do with AYN, something to do with Godot or how it's using SDL. GUIDs are normalized to 32 characters, there is no precedent for shorter, so I'm not going to merge this; the 30+ mappings we have that match this format are all right-padded. Feel free to try the [SDL utils] on Android if you want to unwind this further

harmsc12:
I didn't pay attention to the padding instructions earlier and put the zeroes on the left side.  I just tried it again on the right side and it works.  I'll fix my stuff real quick.

harmsc12:
I feel like a total noob now because of that.

offalynne:
Don't sweat it ! Glad we were able to work out all the kinks :-) Thanks again 👍

PR #759, opened 2024-05-18
Update gamecontrollerdb.txt with Scuf envison-pro

zany130:
And pro model of envision. (when wired, wireless doesn't work) EDIT: got wireless working somehow and added wireless GUID

Map G1 button of both models to misc1 (shows up in Steam as a capture button)


zany130:
Not sure if we want to expose the other "G-KEYS" (extra hot keys on the bottom of the controller)

they are from left to right (so G1,G2,....G5)
btn 11 -15

the profile switch button under the home is btn 16, however, btn 12-15, that is to say, G2-G5 and btn16 (profile switch button) already do stuff even if mapped G2-G4 control volume up, down, and mute. G5 controls eco mode (toogles vibration and RGB) and the profile button obviously controls the profile however you have to double-tap to switch so it still may be helpful to to map this one 

G1 is the only extra button that does nothing by default that's why I mapped it to misc1

offalynne:
Thanks, based on your notes the mappings look good.

PR #763, opened 2024-06-03
Added Raphnet Dual GameCube Adapter v3.6 GNU/Linux mappings

kimimaru4000:
This PR adds mappings for the v3.6 firmware of the Raphnet Dual GameCube USB adapter for GNU/Linux.


offalynne:
Thanks you. Can you confirm face button mapping matches the layout shown in the README ? https://github.com/mdqinc/SDL_GameControllerDB?tab=readme-ov-file#mapping-guide

kimimaru4000:
Thanks @offalynne for making the change and merging! Can I ask why the "Dual" in the name was dropped? Raphnet also makes a single-port GameCube adapter and my intent was to differentiate it. Thanks!

PR #764, opened 2024-06-14
Fixes existing Xbox mapping, adds mapping for new GUID after firmware update

MasonRemaley:
I'm using GLFW's gamepad handling which relies on this database, and noticed that my Xbox controller's mapping wasn't working.

The correct mapping appears to be identical to the mapping used by the majority of GUIDs for Xbox controllers in the database, so I went ahead and updated it to use that.

After updating the mapping, I also updated my controller's firmware to the latest version. There was no mapping for the updated GUID in the database, so I added a mapping for that as well (it's identical, just the GUID changed.)

First time contributor, let me know if it looks like I did anything wrong! :)


offalynne:
Thank you ! Looks good, just missing the `guide` field, if you could add that back to both mappings (`guide:b8,`) all good to go.

MasonRemaley:
It looks like they both already have the `guide` set to `b8`--it's right on the line wrap in my browser so it's easy to miss haha.

PR #765, opened 2024-06-19
Add Microsoft Xbox Series X

Der-Verpruegler:
This provides mapping for Microsoft Xbox Series X controller (via USB and Dongle)

This mapping string was generated with AntiMicroX


offalynne:
AntiMicroX produces GUIDs and mappings incompatible with this db. Please refer to one of the software options in the README.

Der-Verpruegler:
Yes, I adapted (and tested of course) the string that was generated ...

offalynne:
If you have tested both before and after adding the mapping to the DB and confirm behaviour was not correct before, and is after the change, feel free to reopen the PR and I'll merge

PR #779, opened 2024-08-11
Add missing GUIDs for 8BitDo Pro 2 Controller in XInput mode

Zekromaster:
These two are the GUIDs that the 8Bit Do Pro 2 wireless gamepad uses in wireless and wired mode when using XInput. The first one got misidentified as 8Bit Do Zero 2 by some software using the DB, which led me to realise there was some ambiguity which led me to realise this meant the DB didn't contain these specific mappings.


offalynne:
The GUID format is wrong. You need to remove the CRC, by zeroing chars 5-9 so for example `050095ac5e040000e002000003090000` becomes `050000005e040000e002000003090000`.

When correctly formatted, these GUIDs are duplicates and match existing maps in the db.

This has nothing to do with mappings for the 8Bit Do Zero 2. There is no overlap with any mappings that match that device.

This is the third of three bad PRs I have closed regarding this. Please do not open another unless you have confirmed what is wrong, how to fix it, and confirm that your change does in fact correct the problem.

If the issue is downstream software, take it up with maintainers of that software.

Zekromaster:
Oh, didn't realise that. The change did in fact correct the problem, but I assume this just means that the downstream software is not using the mappings correctly.

offalynne:
I assume PCSX2 is parsing the DB incorrectly. Try removing your changes, replacing the mapping [linked above](https://github.com/mdqinc/SDL_GameControllerDB/pull/779#issuecomment-2282803742) with yours, leaving the original GUID alone, and see if with/without that change you can confirm that breaks/fixes the issue.

Zekromaster:
Can confirm this also resolves the issue

Somehow SDL tool gives me a GUID of `050000005e040000e002000003090000`, but other software is giving me the mappings for `050000005e040000e002000030110000` instead.

Zekromaster:
I seem to only be able to actually reproduce any issue when the file contains the following three mappings:
```
050000005e040000e002000030110000,8BitDo Zero 2,a:b0,b:b1,back:b6,dpdown:+a1,dpleft:-a0,dpright:+a0,dpup:-a1,leftshoulder:b4,rightshoulder:b5,start:b7,x:b2,y:b3,platform:Linux,
050000005e040000e002000003090000,Xbox One Controller,a:b0,b:b1,back:b6,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b10,leftshoulder:b4,leftstick:b8,lefttrigger:a2,leftx:a0,lefty:a1,rightshoulder:b5,rightstick:b9,righttrigger:a5,rightx:a3,righty:a4,start:b7,x:b2,y:b3,platform:Linux,
050082795e040000e002000003090000,Xbox One S Controller,a:b0,b:b1,back:b6,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b8,leftshoulder:b4,leftstick:b9,lefttrigger:a2,leftx:a0,lefty:a1,rightshoulder:b5,rightstick:b10,righttrigger:a5,rightx:a3,righty:a4,start:b7,x:b2,y:b3,platform:Linux,
```

This could very well be an issue with SDL, but might be worth asking here before opening an issue there. There seem to be two mappings that are the same GUID, differing only in CRC. It is my understanding from this PR's closing message that this isn't supposed to happen, but maybe it is in certain circumstances?

offalynne:
Correct, the `050082795e040000e002000003090000` entry is in error and will be removed.

> the following three

this is indeed a downstream issue then as (whatever) software you're using is throwing away the revision component of the GUID (the suffix) and only matching on some/all of bus + vendor + product.

if you remove *only* the first and last mappings does it resolve the issue ?

Zekromaster:
Yes, it does. But the downstream software is only setting the SDL hint pointing to the mappings file, so if this is unexpected behaviour the problematic downstream producing it is SDL itself.

PR #783, opened 2024-09-13
add Linux/Windows mapping for Alfadata USB Gameport Adaptor

bomb2777:
What is it?
- a mapping for a USB adaptor (USB to PC Gameport)
- there can be simple 2-axis and 1/2-button controllers connected to this adaptor
image

Why do I need it?
- I am developing a clone of an old game which has only 1 fire button with LibGdx (which uses SDL_GameControllerDB)
- **my USB Adaptor is not recognized** in the game whilst other gamepads like **Logitech F310 (has a mapping in gamecontrollerdb.txt)** is recognized
- **Linux and Windows recognize the adaptor without problems**
- a lot of people have USB adaptors to use old 9-pin (e.g. Amiga) or 15-pin (e.g. PC Gameport) controllers to play (old) games (like the one I clone)


offalynne:
Have you confirmed that LibGdx **_does_** recognize the device after this change?
It is not usual in SDL, the intended consumer of the db, for a generic device to be unavailable simply because a db mapping is not present.

2 axes and 2 buttons are not suitable criteria for a mapping. The goal of mapping is to disambiguate layout; with less than 3 buttons or axes a mapping is redundant and unnecessary.

bomb2777:
> 2 axes and 2 buttons are not suitable criteria for a mapping. The goal of mapping is to disambiguate layout; with less than 3 buttons or axes a mapping is redundant and unnecessary.

thx for your explanation, I will try something else then :+1:

bomb2777:
> Have you confirmed that LibGdx does recognize the device after this change? It is not usual in SDL, the intended consumer of the db, for a generic device to be unavailable simply because a db mapping is not present. If that is the case with LibGdx, you should consider that a bug.

@offalynne JFYI: I got it working with a modified gamecontrollerdb.txt which overrides the orginal one in my game.jar. 

I created a mapping with SDL Gamepad Tool before, but it does only recognize 2 directions, not 4.
Mapping string: `030000009d0700000102000000000000,USB ADAPTOR,a:b1,x:b0,dpup:a1,dpleft:a0,platform:Windows`

I tried to create the mapping with Steam's Big Picture Mode, which produces a more usable output:
Mapping string: `030000009d0700000102000000000000,USB ADAPTOR,a:b1,x:b0,dpleft:-a0,dpright:+a0,dpup:-a1,dpdown:+a1,platform:Windows`

I copied the `dpad` (dpleft. dpright etc.) values to the Linux string, now the USB Adaptor is recognized on both platforms.
Looks not like a bug on libgdx's side for me, but probably the adaptor + controller I use is just too old. 

Have a nice day.

offalynne:
Yes the mapping tool is not always the friendliest, glad you worked out the final correct mapping that works for you

> ***Looks not like a bug on libgdx's side for me***

Just to clarify, SDL should ***not*** require a mapping db entry to for the device to be registered and accessed as a joystick. Given gameport is a generic interface for devices including joysticks, it should "just work" without the mapping you added. Since Steam and the mapping tool both recognize it ***before*** you add the mapping db entry, it seems the issue is with Libgdx. In any case, glad you found your solution!

bomb2777:
Thx again for explanation. I ignored the fact that Steam already uses SDL under the hood. You are right, must be something with the implementation on https://github.com/libgdx/gdx-controllers 's side. Will open an issue there.

bomb2777:
https://github.com/libgdx/gdx-controllers/issues/45

PR #786, opened 2024-09-16
Update gamecontrollerdb.txt

mbabuskov:
Added PS4 Dual Shock connected directly to Debian 11 box via USB cable.


offalynne:
The `16800` suffix indicates this is being handled by HIDAPI making these mappings redundant https://github.com/mdqinc/SDL_GameControllerDB/issues/478

PR #787, opened 2024-09-17
New Linux additions

rtomasa:
Added some new Linux mappings


offalynne:
@rtomasa thanks for your submission (and your work in general !) I'm removing the DragonRise mapping per #202

PR #792, opened 2024-10-04
Add VEX V5 controller

NoozAbooz:


offalynne:
Thank you ! Looking at a photo of the device it seems to match Nintendo iconography for face buttons. Can you confirm your mapping is based on position (eg. south face button is SDL mapping key "a") per https://github.com/mdqinc/SDL_GameControllerDB?tab=readme-ov-file#mapping-guide ?

NoozAbooz:
Sorry, I don't know too much about controller mapping so I'm not too sure what you mean. I did not map it using the letters on the buttons, I just pressed the buttons on my controller that corresponded to the diagram on https://generalarcade.com/gamepadtool/.

PR #798, opened 2024-10-19
Add support for Nacon Revolution 5 Pro on Linux

Billli11:
Is the crc always the 5-8 char in GUID?


offalynne:
> Is the crc always the 5-8 char in GUID?
4 char, removed from GUID before listing, see https://github.com/mdqinc/SDL_GameControllerDB/issues/609#issuecomment-1336432046

PR #800, opened 2024-10-23
Update N64 controller configuration

vanfanel:
This new configuration fits the N64 native game ports expected controls (Mario64, Zelda OOT, etc) by mapping the C buttons to the secondary joystick directions, etc. So this fits the expected default configuration of the N64 pad much better than the previous one.


offalynne:
The SDL mapping for N6 A>B is A>B for sanity's sake.
>  buttons to the secondary joystick directions, etc

The original mapping already does all this.

PR #803, opened 2024-10-31
Add L-Tek Dance Pad mapping

MelissaAutumn:
Device in question: https://www.iamats.com/shop/dance-mat-ex-pro-metal-white/

Mapping was generated from the SDL2_Mapping tool. Normally I'd say this pad isn't worth adding, but SDL unfortunately picks it up via the Gamepad API. Everything works great except for the up arrow, so this mapping fixes that issue.

I've mapped it to A B X Y based on how it was reading from the Joystick API (and then adjusted if needed to map North, South, West, East.) 

The board actually has capacity for a 5th button, and a start and select button (seen https://ddrpad.com/collections/circuit-boards/products/copy-of-l-tek-arduino-dance-pad-shield-1000hz-pcb) but I don't have those hooked up. Let me know if I should manually map those as well.

Let me know if I should adjust the mapping to map to a dpad instead and/or manually add the start/select/5th button. 

Cheers


offalynne:
Thanks for the PR Melissa! 

Unfortunately I don't think this DB is the place for a dance mat mapping to live as we intentionally omit non-gamepad devices. I believe you should take this upstream to the SDL project itself: https://github.com/libsdl-org/SDL/

> *SDL unfortunately picks it up via the Gamepad API*

I believe this can be corrected by assigning the joystick type per vendor/product id. It's not immediately clear to me how non-XInput dance pads are handled, perhaps see [`SDL_JOYSTICK_TYPE_DANCE_PAD`](https://github.com/search?q=repo%3Alibsdl-org%2FSDL%20SDL_JOYSTICK_TYPE_DANCE_PAD&type=code) for starters.

PR #805, opened 2024-11-22
Add Dual Shock 4 revision (Linux)

exkc:
my (Chinese knock-off of) Dual Shock controller has 2 GUID :
`050000004c050000cc09000000810000`
`030000004c050000c405000000810000`

The first one is used when the controller is connected to the linux box via BT and it is in the data base.
The second one is used when the controller is connected to the linux box via USB but it is not in the data base.

I  made a mapper for second one by coping the mapping of the first one.


exkc:
it seem to work . i made a patched of glfw that has included the mapping and my game has recognised the controller.

PR #809, opened 2024-12-01
Update gamecontrollerdb.txt

streamingdv:
Add correct DualSense Edge mappings for Linux


offalynne:
What button indexes do you get for the "Mic" and "Touchpad" buttons ? Any udev tool should work, ideally jstest-gtk

streamingdv:
@offalynne I tried jstest-gtk,  SDL2_Gamepad_Mapper and evtest but none of these tools reacted to any mic button press or touchpad press

Ryochan7:
I can confirm that the DualSense "Mic" and "Touchpad Click" buttons are not registered in evtest.

PR #814, opened 2024-12-25
Update gamecontrollerdb.txt for  Sanwa Supply JY-P76USV (SpeedLink ThunderStrike)

Doronimmo:
Should I add it to the other platforms?


offalynne:
We omit this device per #202

PR #815, opened 2024-12-26
Update gamecontrollerdb.txt with support for Dualsense Edge

eleonore122:
entry added for Dualsense Edge controller on linux


offalynne:
duplicate of L1555

PR #823, opened 2025-01-20
Add EasySMX ESM-9100 mapping for Linux.

thesquaregroot:
Mapping created using Godot Joypads Demo / Tool.


offalynne:
This device is spoofing an Xbox controller, and the GUID looks weird (likely a Godot bug). Can you please confirm the controller ***does not*** work correctly ***without*** this db entry ?

PR #824, opened 2025-01-23
Update gamecontrollerdb.txt

AngeloScipione:
Add for linux:
- Mad Catz V.1 Stick.
- Thrustmaster FGT Rumble 3-in-1


offalynne:
> *Thrustmaster FGT Rumble 3-in-1*

This should be handled as a [wheel device](https://github.com/libsdl-org/SDL/blob/main/src/joystick/SDL_joystick.c#L342) instead

> *Mad Catz V.1 Stick*

This should be handled as a [wheel device](https://github.com/libsdl-org/SDL/blob/main/src/joystick/SDL_joystick.c#L288) instead

AngeloScipione:
Ok... So, I must write code in another project. It's too far of my ability. Would you mind add an advice about not suitable controllers (wheels, etc)? Also It could be helpful add a link to a tutorial for these devices. Thank you very much.

PR #825, opened 2025-01-24
8BitDo SN30 Pro Bluetooth Controller,G Classic,Android/Linux mode

ruyrybeyro:
Yet another entry for 8BitDo SN30, booted with Start+B

8BitDo SN30 Pro Bluetooth Controller, Hall Effect Joystick Update, Compatible with Switch, PC, macOS, Android, Steam Deck & Raspberry Pi (G Classic)


ruyrybeyro:
wrong order.

PR #829, opened 2025-01-26
Flydigi APEX 4 D-input mappings

zany130:
Hi, I created some mappings for the Flydigi APEX 4 controller when in d-input mode (by default home and all the extra buttons are unbound necessitating a custom mapping)

the first one is for Bluetooth and exposes all the paddles, buttons, and extra function buttons as capture.

the second one is for the dongle and only exposes the paddles and normal buttons, the extra function button is read as a right trigger pull in this mode

the third one is for when wired in D-input mode and is the same thing as dongle but with a different ID


offalynne:
Hi, thanks for these. I would like to merge, but in the past SDL has avoided mappings for Flydigi pads because the mappings have changed as frequently as every firmware release. If possible, can you confirm these work with the latest firmware for the controller ?

Thanks again

offalynne:
Thanks for confirming. If you're able to get mappings on Windows (DInput) and/or Mac in the future, please contribute them !

offalynne:
OK, removing CRCs we get

```
05000000b40400001224000001010000,Flydigi APEX 4,a:b0,b:b1,back:b8,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b14,leftshoulder:b4,leftstick:b10,lefttrigger:a5,leftx:a0,lefty:a1,misc1:b20,paddle1:b19,paddle2:b15,paddle3:b16,paddle4:b13,rightshoulder:b5,rightstick:b11,righttrigger:a4,rightx:a2,righty:a3,start:b9,x:b2,y:b3,platform:Linux,
03000000b40400001224000011010000,Flydigi APEX 4,a:b0,b:b1,back:b10,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b15,leftshoulder:b6,leftstick:b12,lefttrigger:a5,leftx:a0,lefty:a1,paddle1:b14,paddle2:b16,paddle3:b17,paddle4:b18,rightshoulder:b7,rightstick:b13,righttrigger:a4,rightx:a2,righty:a3,start:b11,x:b3,y:b4,platform:Linux,
```

`03000000b40400001224000011010000` is already in the db
https://github.com/zany130/SDL_GameControllerDB/blob/76462abdcd8256151812e445302b0b94e2935cb2/gamecontrollerdb.txt#L1269

```

Flydigi Vader 2: paddle1:b16,paddle2:b17,paddle3:b14,paddle4:b15,misc1:b2, 
Flydigi APEX 4:  paddle1:b14,paddle2:b16,paddle3:b17,paddle4:b18,guide:b15,
```

Feel free to make suggestions here. I'm inclined to remove the USB mapping altogether given the conflict

zany130:
you mean the last mapping I added in [`76462ab` (#829)](https://github.com/mdqinc/SDL_GameControllerDB/pull/829/commits/76462abdcd8256151812e445302b0b94e2935cb2)

which is for dinput (USB) is conflicting with the Vader 2 one?

Do they seem to have different IDs to me?

Sorry if I misunderstood. I don't think any of the IDs repeat (they do look extremely similar though :sweat_smile: ) but I made sure to ctrl+f all the IDs I entered to ensure they don't conflict.( I remember reading somewhere here or maybe in the SDL GitHub that has been a pain point with flydigi controllers (in fact the dongle mode identifies as a Vader 3 by default)

I don't have a Mac but I'll try to get some Windows mappings latter (Do mappings even change based on the platform? From what I quickly saw they seem to be mostly the same?)

offalynne:
> *"I don't think any of the IDs repeat"*

Yes, sorry it can be a bit confusing. We have to modify the GUIDs a bit before they make the database.

Your mapping tool gave you the following
`0300f5d4b40400001224000011010000`
`    ^^^^` <- These 4 chars are the CRC

We remove these, so the final GUID conflicts with the Vader 2 mapping.

offalynne:
I don't want to degrade your experience, but I'm inclined to just outright remove any of the conflicting Flydigi mappings that are colliding here. I suggest you instead contribute  upstream to the SDL repo, since they utilize the disambiguating CRCs, where we do not for peace of mind and ease of compatibility.

Sam also has access to some Flydigi controllers himself (I believe) to be able to test directly, as well as a plethora of user data to look at from the Steamworks backend, so he's much better equipped to canonize a mapping with that direct access.

In the past I've outright purchased devices myself to try to make the best of a confusing mapping situation but Flydigi's models number and cost more than would be reasonable for me to do so in this case.

Where a device has not been (testably) identifiable our past decision has been to omit a gamepad mapping so the device can fall through to the general joystick API eg. https://github.com/mdqinc/SDL_GameControllerDB/issues/202

Since most (all ?) of these devices have a generic mode or three anyway, I don't think there's a huge loss lacking a mapping for the DInput mode, even though I agree they would be nice to have.

offalynne:
These changes are the best I can do within our constraints. Note that it leaves the Vader 2 without misc or paddle mappings, since they're in conflict with the Apex 4.

I suspect the Bluetooth Apex 4 mapping (GUID `0500...`) is also in conflict with the Vader 2 and should probably similarly lose the misc and paddle mappings, but since we lack a mapping for the Vader 2 I've left them be.

Note that this leaves the paddles unmapped in this mode. If this is satisfactory, let me know and I'll merge. If not, I'll close this PR and remove the conflicting mappings so the conflicting devices can fall through to the general joystick API.

zany130:
> Yes, sorry it can be a bit confusing. We have to modify the GUIDs a bit before they make the database.
> 
> Your mapping tool gave you the following `0300f5d4b40400001224000011010000` ` ^^^^ ` <- These 4 chars are the CRC

Ah ok got it

> I don't want to degrade your experience, but I'm inclined to just outright remove any of the conflicting Flydigi mappings that are colliding here. I suggest you instead contribute upstream to [the SDL repo](https://github.com/libsdl-org/SDL?rgh-link-date=2025-01-28T08%3A40%3A51Z), since _**they do**_ utilize the disambiguating CRCs, where we do not for peace of mind and ease of compatibility.

Sounds like this may be the best option?



> These changes are the best I can do within our constraints. Note that it leaves the Vader 2 without misc or paddle mappings, since they're in conflict with the Apex 4

Part of me wants to say forget the Vader 2 who uses that one anymore lol, but thinking about it feels like having paddle mappings on the Vader is a better QOL vs. the Apex 4 because the Apex 4 has that screen for easily mapping the paddles on the fly on the controller itself (although this does limit you to only controller buttons)

I can go ahead and move this upstream if you think that's best?

offalynne:
The only GUID difference between the modes is the CRC which is meant to be removed, so for the purpose of this db they are redundant.

> *"I can go ahead and move this upstream if you think that's best?"*

Please, yes

PR #831, opened 2025-02-01
added linux support for the PDP Xbox Atomic controller

Khusheete:


Khusheete:
Hi !

I found that the controller GUID generated by the SDL2 Gamepad Mapper (030086656f0e0000f102000000000000) is not used by other programs. eg. in the Godot game engine, the controller's GUID is 0300000047656e6572696320582d4200.

Is that normal ? Is there a "correct" one ?

offalynne:
Godot does not use SDL per https://github.com/godotengine/godot-docs/pull/10569#issuecomment-2625541786, hence the discrepancy.

PR #837, opened 2025-03-23
Fix `GO-Super Gamepad` mapping

viniciusjarina:
The `190000004b4800000011000000010000` was mapped with two button reversed. I used SDL `controllermap` to regenerate the mapping.


offalynne:
Does the mapping match position -- not label, as per the README ?
https://github.com/mdqinc/SDL_GameControllerDB?tab=readme-ov-file#mapping-guide

viniciusjarina:
Correct. Previously was matching the label (which is wrong)

Issue #844, opened 2025-04-16
Stadia Controller analog sticks and triggers are mis-mapped


Eggmanplant:
Read the title. Likely the Stadia Controller firmware got updated and changed the maps. My mappings:
    - dpup: h0.1
    - dpdown: h0.4
    - dpleft: h0.8
    - dpright: h0.2
    - y: b3
    - a: b0
    - x: b2
    - b: b1
    - back: b6
    - start: b7
    - leftshoulder: b4
    - rightshoulder: b5
    - leftstick: b9
    - rightstick: b10
    - guide: b8
    - misc1: b11
    - misc2: b12
    - lefttrigger: a10
    - righttrigger: a9
    - leftx: a0
    - lefty: a1
    - rightx: a2
    - righty: a5
    Also added the screenshot and mic buttons as misc1 and misc2.

offalynne:
how is the device connected
what platform is this
what is the GUID sting

offalynne:
I can only assume you used a registry patch to rewire DirectInput axes in which case your mappings are invalid. If you confirm your registry is unmodified (or are not using Windows) open a PR with the full mapping including platform and GUID

PR #845, opened 2025-04-17
Probably fix Stadia Controller on Linux

Eggmanplant:
Read the title. I don't know how to letter sort buttons. It's probably because Google made specific Linux drivers in 2024.


offalynne:
Fixes #844

offalynne:
Can you post more information about your setup: distribution, software, etc. please ?
Testing on Ubuntu and SteamOS and the mappings I'm seeing match those that are in the db, yours do not work correctly.

PR #847, opened 2025-04-18
add a third-party GC controller to USB adapter

Azure-Lazuline:
i hope i'm doing this right! I have this old WiiU/PC gamecube adapter that wasn't in the database.


offalynne:
`6800` indicates it should already be working. https://github.com/mdqinc/SDL_GameControllerDB/issues/478
is it?

offalynne:
With the HID driver this mapping shouldn't be necessary, without it there's a match here https://github.com/mdqinc/SDL_GameControllerDB/blob/3266af403894a437fbee282ffd0defbfdd767f25/gamecontrollerdb.txt#L920

Azure-Lazuline:
oh, to clarify, it *was* reading the controller but the B and X buttons were swapped (matching the letters rather than the positions, so it had B on the right and X on the left), so i made the mapping to fix that. Everything else was fine. I had checked using SDL3 testcontroller after noticing the swap in another game. Is it anything wrong with my setup? I'm kinda confused what part of the pipeline the error's coming from then.

offalynne:
Thanks for clarifying, and no, no issue on your end. The intention is definitely for B/X to match on position rather than label; that's definitely an upstream issue you should take to SDL: https://github.com/libsdl-org/SDL

Squall-Leonhart:
> Thanks for clarifying, and no, no issue on your end. The intention is for B/X to match on position rather than label [per the mapping guide](https://github.com/mdqinc/SDL_GameControllerDB?tab=readme-ov-file#mapping-guide) so that's an upstream issue you should take up with the SDL project directly: https://github.com/libsdl-org/SDL
> 
> See [`USB_PRODUCT_EVORETRO_GAMECUBE_ADAPTER2`](https://github.com/search?q=repo%3Alibsdl-org%2FSDL%20USB_PRODUCT_EVORETRO_GAMECUBE_ADAPTER2&type=code)

SDL doesn't use HIDAPI for these adapters unless built with libusb and the zadig driver is presently installed on the adapter.

offalynne:
@Squall-Leonhart the user contributed data indicates their device was being handled by HIDAPI. https://github.com/mdqinc/SDL_GameControllerDB/issues/478

PR #848, opened 2025-04-18
Change NSO N64 Controller mappings from hid-nx-dkms to upstream hid-nintendo

0B001C:
The original mappings for the NSO N64 controller were created for [hid-nx-dkms](https://github.com/emilyst/hid-nx-dkms), an out-of-tree fork of the mainline [hid-nintendo](https://github.com/torvalds/linux/blob/master/drivers/hid/hid-nintendo.c) driver.

hid-nintendo has since gained support for the NSO controllers in Linux 6.8 while hid-nx-dkms was last updated for 6.1 and seems unmaintained. [This open PR](https://github.com/emilyst/hid-nx-dkms/pull/23) suggests that it doesn't even build against kernels >= 6.12 anymore.

Unfortunately the way the buttons are mapped to evdev events is different between these two drivers, requiring an update to the SDL mappings.

As I understand it this results in a breaking change since the GUIDs are the same. This change therefor trades of support for a legacy out-of-tree DKMS module against proper out-of-the-box mappings on modern kernels.

This PR only addresses the N64 controller. The NSO SFC / SNES and Mega Drive / Genesis Controllers are also affected but I don't have access to the hardware to verify any new mappings. The NSO Famicom / NES Controllers should be fine as-is.


offalynne:
> *This change therefor trades of support for a legacy [..] module against [..] mappings on modern kernels*

Thanks, this is always our preference. I will test with the SFC Online pad, as well.

offalynne:
Present SFC NSO pad mappings seems good with hid-nintendo

offalynne:
I suspect I added the mappings for the SFC NSO pad post hid-nintendo ? In any case, seems to work fine

offalynne:
Not sure why the dkms mappings are working for me, perhaps I need to manually remove the driver

A>A, B>B is correct for N64, thanks for following the format

offalynne:
@0B001C Can you confirm you actually saw `050000007e0500001920000001000000` as a GUID under this driver and the mapping matches, or did you copy the mapping across?

As I understand it this is the generic GUID and mapping seen through evdev without a driver and thus should not use the updated driver mapping

offalynne:
e7aecb6af5d43011c2f514e917eee588964ffa66

PR #855, opened 2025-05-29
Add Xbox Series X Controller (Linux - Wireless)

b1bradders:
I have no idea why but GLFW would not detect my controller's mapping, despite the SDL2 Gamepad Tool being able to. I came here to try to find the mappings for my controller to add them to GLFW's `mappings.h`, but I instead found nothing so here I am now.

The mappings themselves are both the same, and came from the one that the SDL2 Gamepad Tool already detected, but the different IDs come from:
1. SDL2 Gamepad Tool - `0500509d5e040000130b000022050000`
2. What GLFW reported when doing `glfwGetJoystickGUID(0)` - `050000005e040000130b000022050000`

<sub>The difference is the GLFW GUID missing `509d` after the first 4 digits.</sub>

Perhaps the Series X mappings were purged at a later date, after SDL2 Gamepad Tool v1.1.2 was built, which is why the SDL2 Gamepad Tool v1.1.2 can still detect the mapping, but I can't find it anywhere (in this .txt file).

Hoping this helps some other confused soul out there with this controller's GUID lmao.


StupidRepo:
Hang on - I think I need to clone the mappings for when it's wired too, as the name is different compared to when it's wireless.

StupidRepo:
@offalynne Thank you!
<sub>What's CRC mean though?

offalynne:
https://github.com/mdqinc/SDL_GameControllerDB/issues/609

PR #860, opened 2025-06-14
Add Hori Pad A

wantija:
Adding Hori Pad A, this is for the iDOLM@STER Cinderella Girls themed Hori PS3 gamepad.


offalynne:
This is a duplicate of `Hori Gem Pad 3` mapping immediately above (with CRC added) so I'm removing it

PR #868, opened 2025-07-23
Added 8BitDo N30

munsie:
This is a mapping for the 8BitDo N30 (https://www.8bitdo.com/n30-wireless-for-original-nes/) when connected via USB.  Tested with SDL3 on macOS 15.5 on MacBook Pro M4 Max.


offalynne:
Not going to accept this unfortunately as it's spoofing an Xbox controller in this mode, and we want to avoid conflicts with original hardware when matching VID/PID

PR #872, opened 2025-08-12
feat: add mapping for NSO NES Controller on linux

bmsuseluda:
This change adds a mapping for the NSO NES Controller on Linux. On Windows this is not necessary.


bmsuseluda:
interesting, because it does not work for me without them🙁
is there something i need to do with sdl2 without the 4 CRC chars?

offalynne:
If you are able to double check your results on the version of SDL you're using and the issue persists, I suggest opening an issue with the SDL project and coming back if indeed you're able to work out a mapping-db related solution: https://github.com/libsdl-org/SDL/issues

PR #875, opened 2025-08-12
feat: add mapping for NSO N64 Controller on linux and windows

bmsuseluda:
This change adds mappings for the NSO N64 Controller on Linux and Windows.


offalynne:
Duplicate per https://github.com/mdqinc/SDL_GameControllerDB/pull/872#issuecomment-3180339766

PR #893, opened 2025-10-24
Fix source link in gamecontrollerdb.txt

TriangulumDesire:
This is a micro-PR that fixes the source GitHub link in `gamecontrollerdb.txt`, as it hasn't been updated since this repository was moved to the *mdqinc* organisation.

GitHub still redirects the old link, so it's no major issue, but I figured it'd be best to update it.


offalynne:
Thank you

PR #894, opened 2025-11-19
Add G-Shark GS-GP702 controller mapping

coco875:


offalynne:
This device is intentionally omitted, see #202

PR #902, opened 2025-12-11
add icedragon.io stac1/2 mappings

dinsfire64:
I sell dance pad io boards which have some issues being mapped 1:1 to an xinput type device, especially when the button inputs are higher it might be interpreted by a rutter or axis by certain game engines.

This maps the arrows to face buttons (because the controller needs to be able to recognize simultaneous opposing dpad inputs like up&down or left&right) just so games like older windows games under wine can easily pick them up.

each player side (set by a DIP switch) has a guaranteed VID/PID combo to ensure that the guids by sdl should be unique for mapping purposes.

product page: https://icedragon.io/stac2/


offalynne:
Do you have a compelling reason to include these in the gamepaddb ? Would they not already work acceptably out-of-the-box with whatever downstream software might use them in absence of the mappings ?

Asking given the db is intended for gamepads
https://github.com/mdqinc/SDL_GameControllerDB/blob/ff1dc973268613a6fcc4b5017f5537c845e0bfe8/README.md?plain=1#L46-L47

And SDL does seem to support dance pads as separate subset type of joysticks, as gamepads are
https://github.com/search?q=repo%3Alibsdl-org%2FSDL+dance_pad

Not necessarily opposed to the contents of the PR, curious mostly

offalynne:
Future PRs for any of Windows, MacOS, or Android mappings would be welcome

PR #906, opened 2025-12-29
Add missing Raphnet N64 Adapter v3 Linux mapping

Francessco121:
Adds a missing Linux mapping for the v3 Raphnet N64 Adapter. The Windows mapping for the same exact adapter (`030000009b2800006100000000000000`) already exists in the database.


offalynne:
No problemo, we remove the CRC from the GUID so
`030032039b2800006100000001010000`
becomes
`030000009b2800006100000001010000`
for the DB, fwiw.

Feel free to contribute in the future if you find anything else missing 🤙

PR #922, opened 2026-02-14
Add EasySMX X20 to gamecontrollerdb.txt

Chuli2k:
Added the EasySMX X20 controlled config for Linux

PR #923, opened 2026-02-16
Add easysmx x20

Chuli2k:
Creating a new PR as I couldn't reopen the other one #922.
I removed the CRC and checked the triggers and they are ok. They are analog.

For reference this is what `testcontroller` gave me as output:
`03008519790000001c18000011010000,EasySMX X20,a:b0,b:b1,x:b3,y:b4,back:b10,guide:b15,start:b11,leftstick:b13,rightstick:b14,leftshoulder:b6,rightshoulder:b7,dpup:h0.1,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,leftx:a0,lefty:a1,rightx:a2,righty:a3,lefttrigger:a4,righttrigger:a5,platform:Linux,crc:1985,`

As for "TGZ Controller" having the same GUID, I don't know how to handle that. Maybe it'll be faster if you add the config :P


offalynne:
If you are able, can you post a couple screenshots from jstest indicating the trigger held down, eg. left trigger (image) right trigger (image) as per below

PR #942, opened 2026-04-14
Add support for RetroBit USB Sega Saturn Controller in its default mode (Linux)

twoda5h:


offalynne:
https://github.com/mdqinc/SDL_GameControllerDB/blob/e6f9f10b2616badca4849e2d8ac1fa175114d854/gamecontrollerdb.txt#L1728

twoda5h:
the GUID i committed isn't on the controller list. it is incredibly likely this controller has SEVERAL different revisions.

twoda5h:
That controller:
> 03000000790000001100000011010000

My controller (as reported in SDL2):
> 0300c6b1790000001100000011010000

offalynne:
https://github.com/mdqinc/SDL_GameControllerDB/issues/609#issuecomment-1336432046
> *always clear the CRC from the GUID*

The GUID in your PR already matches an entry in the DB as linked above. Your PR has literally no effect and is a duplication of data already in the DB.

Do not open further PRs without testing to confirm what you are submitting actually does something.

https://github.com/mdqinc/SDL_GameControllerDB/blob/c6062ec4b92254aced84843d94d379410774d907/README.md?plain=1#L44-L48

Emphasis on the last point.

PR #945, opened 2026-04-24
:wrench: Adds Elite Series 2 for MacOS

ekwoka:
Adds Xbox Elite Series 2 to MacOS configs.

Note: Bluetooth shows it as `Xbox Wireless Controller` but seems like the more specific name would be beneficial.

(Not really sure how to test this for my purposes in Bevy yet)


ekwoka:
Tried to test it but I can't tell if it's not working or the issue is something else...

offalynne:
> Not really sure how to test

> I can't tell if it's not working

Please refrain from opening PRs before testing to confirm the validity of your changes.

https://github.com/mdqinc/SDL_GameControllerDB#contributing

ekwoka:
The problem there is that it's not so easy to test.

I think there still is an issue with this part, but there's also just plain other issues with dealing with objc interfaces.

PR #955, opened 2026-06-06
Added SteelSeries Nimbus Cloud

Danik1601:
So [SteelSeries Nimbus Cloud](https://steelseries.com/gaming-controllers/nimbus?model=cloud) is an Android and iOS controller that somehow works natively on those platforms in both extended shell wired mode and Bluetooth mode.

The other platfrms are supported only in Bluetooth mode. There is a mention of PC among supported platforms, however in my experience that turned out to be false.

I used [Gamepad Tool](https://generalarcade.com/gamepadtool/) to generate SDL string for all three desktop platforms (Windows, macOS, Linux). I have no idea how to do the same on Android and iOS, so no mappings for them. That software showed `mapping available` out of the box on Linux and macOS, but not on Windows. However, the controller didn't work until it was mapped manually.

This gamepad has two back buttons, they are mapped as `paddle1` and `paddle2`.


Danik1601:
The trigger are mapped on Windows as `b` (button) and not an `a` (analog). On macOS and Linux they are mapped as analog. I have no idea why

PR #981, opened 2026-08-27
Add Padix Rockfire MaxFire G-08XU Gamepad

C0rn3j:
* Add `Padix Rockfire` `MaxFire G-08XU Gamepad` Linux mapping
* Bump CI action version and fix indentation
* Fix readme macOS styling

The vendor in the USB data is actually `Padix (Rockfire)` but I changed it to look like the only other existing padix entry - should I have not? It also seems to be distributed by Genius, and the physical label doesn't say jack about Padix/Rockfire.
https://linux-hardware.org/?id=usb:0583-a000
 
I used the SDL2 gamepad tool, moved `platform:*` to the end and deleted the `crc:ece3` from the entry  and GUID as `crc` does not seem to be present anywhere else.

What is the point of specifying platform?

It seems that some controllers (randomly checked `8BitDo GameCube`) share the mappings between OSs, why would they diverge at some point? Whatever the answer, it would be useful to add it to the readme.

This is a horrible controller by the way:

image

image


offalynne:
https://github.com/mdqinc/SDL_GameControllerDB/commit/b0c976e384da0babca93de8807217550b1e5ba2e

PR #987, opened 2026-09-03
Fix inverted left/right paddle mapping for Flydigi Direwolf 4

itsHanibee:
Swaps paddle1 (right) and paddle2 (left) so M1 and M2 trigger the correct sides. Updating kernel PR accordingly.
Functionally doesn't change anything, just an oversight when I was using the GUI mapping tool


offalynne:
> *kernel PR*

??

itsHanibee:
rn the controller only supports the basic xbox360 functionality by default, i'm pushing a PR to xpad fixing that and adding the paddle support.

offalynne:
> *pushing a PR to xpad*

OK, I appreciate the mapping is in service of that kernel patch, but please consider the fallback situation (if possible) and test compatibility with pre-patch behaviour as well as xpad noop, if possible.

In the past kernel drivers have opted to modify the version or some other field that GUID composition uses which helps ensure devices **_not_** using the driver will not be assigned an incorrect mapping.

itsHanibee:
thank u for merging, on unpatched kernels the controller just shows up as `045e:028e` (generic Xbox 360 pad), SDL assigns it the standard Microsoft Xbox 360 GUID, which safely falls back to standard Xbox controls without touching this mapping. on the patched driver the driver overrides the input device ID to the actual hardware ID (37d7:2414) and exposes the paddle buttons.
Because the `03000000d73700001424000004010000` GUID is only generated when the patched driver is active, unpatched systems will never match this entry or receive an incorrect mapping.

i hope to send over the mapping for windows and mac os soon too!

Issue #984, opened 2026-08-31
Incorrect face button mapping for PS3 Controller 03000000632500007505000000000000


blat-blatnik:
A and B, and X and Y buttons are swapped. They used to be correct before, then someone "fixed" them in commit fa9c1fb. Maybe there are two different controllers with the same GUID? Either way, I have this controller and the new mapping is wrong for it.

<img width="480" alt="Image" src="https://github.com/user-attachments/assets/6a3cc6c5-ed01-4402-ab50-be68c913af1e" />

offalynne:
The new mapping is from a Retro Fighters Defender Pro rather than whatever the lovely specimen you've shared with us here is. 

> *Maybe there are two different controllers with the same GUID?*

Correct

If you would like the change reverted, feel free to open a PR.

PR #989, opened 2026-09-04
Fix Qware PS3 Controller face button mapping

blat-blatnik:
Fixes incorrect mapping I mentioned in issue #984.

Apparently multiple controllers have the exact same GUID, so I have no idea how this can ever be resolved in a satisfying manner, but here it is.


offalynne:
Thank you for the PR

> I have no idea how this can ever be resolved in a satisfying manner

It's maintainers' discretion -- usually I try to defer to the higher market share device when possible using available data in case of conflicting mappings. [Upstream SDL](https://github.com/libsdl-org/SDL) has finer tools for this including a CRC field which is meant to differentiate devices. We omit this in the community DB in favor of simplicity and back compat.

